PR #49, opened 2020-08-12
Extract zonation from grid

tnatt:


src/fmu/tools/qcforward/extract_grid_zone_tops.py
```diff
@@ -0,0 +1,133 @@
+import pandas as pd
+import xtgeo
+
+
+class MockProject:
```
berland, line 5: Belongs in the test suite?

PR #53, opened 2020-08-26
Qcforward add perf

jcrivenaes:
Many changes, structural changes

PR #54, opened 2020-08-27
Drop py27, blacken with black 20.8b0. Redo black install in CI.

berland:


.github/workflows/fmu-tools.yml
```diff
@@ -29,28 +29,19 @@ jobs:
...
 
-    - name: Install numpy for Python 2
-      if: matrix.python-version == '2.7'
+    - name: Install requirements
       run: |
-        pip install numpy==1.16.6
+        pip install pip -U
+        pip install git+https://github.com/equinor/libecl
+        pip install git+https://github.com/equinor/fmu-ensemble
+        pip install .[tests]
```
jcrivenaes, line 37: What do `pip install .[tests]` do?

berland: it is the same as `pip install .` but it also picks up "extras" for `tests`as declared in `setup.py` and installs those dependencies also.

jcrivenaes:
Most of this is also done in #53. Think I will align these PR's and just merge #53 to master if OK

.github/workflows/fmu-tools.yml
jcrivenaes: Nice: new to me

jcrivenaes:
Implemented in #53

PR #57, opened 2020-09-01
Extract property statistics

tnatt:
Functionality to extract property statistics from wells, blocked wells and grid properties.

- [x] Classes to extract statistics

- [x] QCforward function for grid statistics

- [x] Tests

- [x] Documentation

- [ ] Add JSON schema validation


HansKallekleiv (commented):
There is a lot of functionality here, so this will take some time to review.
Just some general comments/questions now from a user perspective inside RMS.

- The documentation is good and the syntax is easy to understand
- Did a test on Reek to compare wells, bwells and geogrid. The possibility to rename "selectors" and individual code names makes it possible to compare statistics for all the datatypes without scripting :+1:  

- The blocked wells (and probably wells?) should not require a list of well names, i.e. use all wells by default ?

- Suggest adding an example yaml file using e.g. Reek? to the /examples folder

src/fmu/tools/qcforward/_qcforward_data.py
```diff
@@ -271,30 +291,98 @@ def read_wells(self):  # pylint: disable=too-many-branches
...
+                        mywell.from_file(wellentry, lognames=self._wells_lognames)
+                        wdata.append(mywell)
+                        CMN.print_debug(wellentry)
+
+        else:
+
+            # roxar API input:
+            # wells: {"names": WELLS, "bwname": BW, "grid": GRID}
+
+            wnames = self._data["bwells"].get("names", None)
```
HansKallekleiv, line 360: Set to all wells instead of `None` ?

src/fmu/tools/qcproperties/_combine_propstats.py
```diff
@@ -0,0 +1,61 @@
+import pandas as pd
+
```
HansKallekleiv, line 2: It is a bit hard to understand what `self` is here.
Consider either putting it in the same file as the QCProperties class or rewrite functions to not take in self ?

src/fmu/tools/qcproperties/_create_property_df.py
```diff
@@ -0,0 +1,98 @@
+import pandas as pd
```
HansKallekleiv, line 1: Same as above.

src/fmu/tools/qcproperties/qcproperties.py
```diff
@@ -0,0 +1,242 @@
...
+                fine grained control, e.g. ["grid", "gridprops", "wells"]
+            project (obj or str): For usage inside RMS
+
+        Returns:
+            A PropStat() instance
+
+    """
+
+    def __init__(self):
+        self._gdata = None  # QCForwardData instance, general data
```
HansKallekleiv, line 48: The `QCProperties` class should be independent of the `QCForward` class, but currently the data loading is done using `QCForwardData`.

Perhaps we should consider moving the data loading out of the `QCForward` code base?

Thoughts @jcrivenaes ?

tests/data/propstatistics/OP_1.bw
```diff
@@ -0,0 +1,69 @@
+1.0
```
HansKallekleiv, line 1: Move this file to xtgeo-testdata.

jcrivenaes (approved):
Think we just merge it, and then continue with "code tuning" and refactor afterwards

PR #63, opened 2020-10-07
Add support for discrete properties in qcproperties

tnatt:
Added support for calculating percentages for discrete properties in QCProperties. 

- [x]  Support both regular Percent and a Weighted Percent.

- [x] Minor adjustment of QCForward method which checks grid statistics to ensure it can be used to also check percentages for discrete properties.

- [x] Some bug fixes in both QCForward and QCProperties

- [x]  Removed file _property_dataframe.py and placed the functions within the PropStat class instead.

- [x]  Function in QCProperties to extract weighted average continous properties is removed. It is now computed together with the other "aggeregations". 

- [x] Cleaned up tests for QCProperties, and fixed warning message that there had been several exception when running pytest.

- [x]  Updated documentation

- [ ]  Add some more tests for discrete properties in tests

- [ ]  Add support for getting bulk parameter from XTGeo when functionality is in place


jcrivenaes (approved):
Looks good but many changes. I think we do as before, merge as-is and iterate afterwards

jcrivenaes:
Why are not the github-actions checks ran? Is it required that the check-list in the comment in top is fullfilled?

tnatt:
> Why are not the github-actions checks ran? Is it required that the check-list in the comment in top is fullfilled?

Strange will look into why the actions didn't run? Regarding the check-list, I might throw in a couple of more tests today, but the XTGeo bulk calculation can come in a separate PR when functionality is in place in XTGeo 👍

PR #64, opened 2020-10-13
Support having multiple properties in one file in QCProperties

tnatt:
Added an optional 'pfile' argument for parameters. Needed if multiple properties are located in one file e.g. for reading INIT files.


HansKallekleiv (approved):
Minor suggestion, otherwise LGTM :+1:

PR #67, opened 2020-11-01
Remove dependency to old webviz

HansKallekleiv:
Remove code that generates tornadoplot for [the old webviz module](https://github.com/equinor/webviz), as this is no longer maintained.

The relevant code has been transferred to the [new webviz framework](https://github.com/equinor/webviz-subsurface/).


tralsos:
Yes, I agree we should remove the _add_webviz_tornado_onebyone, the corresponding yaml config files and _find_combinations. However I would like to keep _tornado_onebyone as this is still useful for sensitivities for other simulated responses than currently covered by webviz.

HansKallekleiv:
@tralsos  Can you have a look over now and see that the relevant functionality is kept?

docs/fmudesign.rst
```diff
@@ -315,43 +236,3 @@ monte carlo sensitivities.
...
-realization 0. Statistics showing the difference to the reference can
-be calculated as absolute values, or as percentages. You could also
-choose to exclude from the plot, sensitivities that are smaller than
-the seed sensitivity P10/P90.
-
-.. code-block:: python
-
-    #!/usr/bin/env python
-    # -*- coding: utf-8 -*-
-
```
tralsos, line 337: This example belongs to the code that will be remaining in repository and can be kept: the rest can be deleted.

tralsos: I think the rest of the changes looks ok.

HansKallekleiv: like so?

tralsos (approved):
Yes, looks good now. Thanks for doing this.

berland (approved):

PR #69, opened 2020-11-18
create qcset2zero.rst file with demo and doc to discuss it

JiZur:


jcrivenaes:
The commited file is not on Sphinx rst format, so you need to reformat it. Look at the other *.rst files in docs.

Next I think the doc shall focus on a simple and still flexible "client side design" for the user to _empty_ the icons in RMS, which is not the same as setting values to zero. There are functions for that in the Roxar API for various data types.

For instance, I would propose the client script to be like this:

```
from fmu.tools.rms import setempty

SETUP = {
      "horizons_categories": ["DS_xx1", "DS_xx2"],
       "zones_categories": ["IS_yyy1"],
       "grid_model": {
                  "Geogrid": ["PHIT", "Regions"],
                   "Simgrid": ["PERMX", "PERMY", "PERMZ"]
        },
}
  
def main():
     """Main routine for emptying icons."""
     setempty(SETUP)

if __name__ == "__main__":
     main()

```
This is just a proposal which may need iteration. When the doc is matured a bit, coding can start.

JiZur:
Make some doc / .rst file with (hopefully proper format). 

Code-wise, I suggested to use empty list so that the code try to empty or change prop values to all available data for the given horizon_category / zone_category / grid model in order to avoid long lists of properties when the user wants to apply the change to them all

docs/qcreset.rst
```diff
@@ -0,0 +1,218 @@
+The qcreset module
+=====================
```
berland, line 2: `pip install rstcheck` and then run `rstcheck` on this file will give you some details to fix :)

(I will add `rstcheck`to the CI pipeline eventually)

docs/qcreset.rst
```diff
@@ -0,0 +1,218 @@
...
+        "project": project,
+        "horizons": {
+            "horizon_category_1": ["horizon_name_1", "horizon_name_2"],
+            "horizon_category_2": [
+                "horizon_name_1",
+                "horizon_name_2",
+                "horizon_name_3",
+                "horizon_name_4"
+            ],
+            "horizon_category_3": []
```
tnatt, line 84: Instead of having an empty list, maybe switch to "all" as it is easier to understand.

docs/qcreset.rst
```diff
@@ -0,0 +1,218 @@
...
+^^^^^^^^^
+
+.. code-block:: python
+
+    from fmu.tools import qcreset
+    import roxar
+
+    SETUP = {
+        "project": project,
+        "horizons": {
```
tnatt, line 76: I guess the average user will mostly empty an entire horizon folder or an entire grid, so maybe we should support giving in a list as well for usability? e.g "horizons": ["horizon_folder1", "horizon_folder2"]

docs/qcreset.rst
```diff
@@ -0,0 +1,218 @@
...
+        },
+        "value": -5.0
+    }
+
+    SETUP2 = {
+        "project": project,
+        "grid_models": {
+            "Geogrid": ["discrete_property_1", "discrete_property_2"]
+        },
+        "value": 999.0
```
jcrivenaes, line 213: Giving discrete props a float value 999.0 seems strange

docs/qcreset.rst
jcrivenaes: Agree, and empty list looks like "None"

docs/qcreset.rst
```diff
@@ -0,0 +1,218 @@
...
+
+Signature
+~~~~~~~~~
+
+The input of this method is a Python dictionary with defined keys. The key
+"project" is required while "horizons", "zones" and "grid_models" are optional
+(at least one of them should be provided for the method to have any effect).
+
+project
+  The roxar magic keyword ``project`` refering to the current RMS project.
```
jcrivenaes, line 35: Since this is only inside RMS, "project" is not needed in the arguments list

docs/qcreset.rst
```diff
@@ -0,0 +1,218 @@
...
+Example 1
+^^^^^^^^^
+
+.. code-block:: python
+
+    from fmu.tools import qcreset
+    import roxar
+
+    SETUP = {
+        "project": project,
```
jcrivenaes, line 75: Skip project

docs/qcreset.rst
JiZur: agreed, I was thinking 999 as integer, will update

docs/qcreset.rst
JiZur: But if project is only called in the method (stored on github), you will get undefined variable pylint error isn't it ? Isn'it more elegant to parse it as argument instead of creating exception? What's your opinion on that ?

JiZur: Also, I guess the method needs to be available in RMS python in order to work, not komodo, isn't it ?

docs/qcreset.rst
JiZur: should we use keyword like "ALL" (without any [ ] around) ?

docs/qcreset.rst
JiZur: @jcrivenaes :  no opinion on that topic ?

jcrivenaes: Yes this must run inside RMS (or run in RMS licensed python shell). The issue with pylint complaining on "project" cannot be avoided. Of course you may think that at some stage the project is not the magic but a RMS project reference (if ran from terminal in a RMS python python shell), so... keep it for now and then we can evaluate when testing.

docs/qcreset.rst
jcrivenaes: Yes. I prefer small letters, "all"

JiZur:
Any new feedback on the latest doc ? (I made the code for the set_data_constant() method, I'm planning to push it after we validate this PR)

@jcrivenaes any clue what's the cause of the failure while testing the Sphinx doc building ?

JiZur:
error message is:

checking consistency... failed
/home/runner/work/fmu-tools/fmu-tools/.eggs/setuptools_scm-4.1.2-py3.6.egg/setuptools_scm/git.py:68: UserWarning: "/home/runner/work/fmu-tools/fmu-tools" is shallow and may cause errors
  warnings.warn('"{}" is shallow and may cause errors'.format(wd.path))

Warning, treated as error:
/home/runner/work/fmu-tools/fmu-tools/docs/qcreset.rst:document isn't included in any toctree
Error: Process completed with exit code 1.

berland:
Try to add your rst file to `index.rst`

JiZur:
anyone knows why I'm still failling some tests in files that are inherited from master ?

jcrivenaes:
Something changed outside our control, the ```xlrd``` package: https://pypi.org/project/xlrd/ : _This library will no longer read anything other than .xls files_. I will try to adjust requirements. @tralsos this used by design matrix.

JiZur:
Perfect, thanks Jan ! Anyone having some more feedbacks regarding the documentation ? If not we can approve it, and then I will submit the actual functions with the next PR ?

jcrivenaes (approved):
Looks good now

PR #70, opened 2020-11-19
Refactoring create_ert_rft_files

berland:
* [x] Refactor main script to allow testing (putting everything into `main()?`
* [x]  Establish mocked RMS project that can give out the well coordinates in Drogon
* [x] Allow tests on ad-hoc data sets (though the mocked project)
* [x] Redo user configuration.
* [x] Pylint score 9.94
* [ ] Pylint score 10.0
* [x] Unit tests for all functions
* [x] RST documentation

Proposal to only have this inside RMS:
```
from fmu.tools import create_ert_rft_files

CONFIG = {
  "inputfile": "foo.csv"
  ...
}
create_ert_rft_files.main(CONFIG) 
```

@JiZur , you might want to jump in here..


berland:
Propose to change name to "create_rft_ertobs", to be in line with https://github.com/equinor/subscript/blob/master/src/subscript/merge_rft_ertobs/merge_rft_ertobs.py

berland:
Added export of a CSV files in  `exportdir` called `rft_ertobs.csv`, that contains all the data in the welldatefile, txt and obs files. The intention is to allow a future GENDATA_RFT job to pick up this instead of parsing all the other files.  This will also allow dropping merge_rft_ertobs and read this CSV directly into Webviz.

rnyb (approved):
doc looks good

berland:
@akia, @jcrivenaes , @JiZur anyone has  a chance to look at the code and do some review?

anders-kiaer (approved):
LGTM.

jcrivenaes (approved):
OK. But I see we need to align a naming convention how to organize tests vs data. E.g. we have `tests/test_qcforward` folder with tests; we have `tests/data` with data, but `tests/rms` contains both tests and data. Somewhat confusing.

berland:
> OK. But I see we need to align a naming convention how to organize tests vs data. E.g. we have `tests/test_qcforward` folder with tests; we have `tests/data` with data, but `tests/rms` contains both tests and data. Somewhat confusing.

#87

PR #80, opened 2020-11-30
Add second -f option to git clean

HansKallekleiv:
--force has to be called twice as per https://git-scm.com/docs/git-clean#Documentation/git-clean.txt---force

;)


berland:
Included it in #81

PR #86, opened 2020-12-03
Add badges, use RST format for consistency with sibling repos

berland:


jcrivenaes (approved):
You can mix MD and RST by using recommonmark, but no big issue, so OK

berland:
Yes, it was the reuse of badge code from other projects that made me switch.

PR #91, opened 2020-12-16
Qcreset set data constant

JiZur:
PR request with a first version of the method setting data values to a constant. Please provide feedbacks on the code :)


src/fmu/tools/rms/set_data_constant.py
```diff
@@ -0,0 +1,205 @@
...
+                            "keyword '" + horizons + "' not recognized, 'all' expected!"
+                        )
+                elif isinstance(horizons, list):
+                    for horizon in horizons:
+                        set_safe_value(project, "horizon", horizon, data_type, value)
+
+        else:
+            assert False, (
+                "Value associated with key 'horizons' must be of " "type list or dict!"
+            )
```
tnatt, line 121: This section is very similar to the one below for zones, and could be in a separate function.
The only difference is use of `project.horizons, project.zones`, which can be retrieved as `getattr(project, <category>)`, where category is e.g. zones or horizons.

tnatt (changes requested):

JiZur:
@jcrivenaes , @tnatt branch updated with master, if any of you have time to review :)

HansKallekleiv (changes requested):
LGTM, except linting 👍

JiZur:
@HansKallekleiv it should be ok now... only thing is it's complaining about "import roxar" not being visibly used (but still used to read the project)

src/fmu/tools/rms/set_data_constant.py
```diff
@@ -0,0 +1,196 @@
...
+            of the grid models where properties need to be made empty. The
+            value associated to this key should be a list of property names to
+            modify. If a string ``all`` is assigned instead of a list, all
+            available properties for this grid model name will be used.
+            Alternatively, if a list of grid models names is given instead of a
+            dictionary, the method will apply to all properties within these
+            grid models.
+    """
+
+    import roxar
```
HansKallekleiv, line 54: Why is this import needed?

HansKallekleiv (changes requested):
Documentation is missing? I see it was in some of the earlier commits, but  it has been removed.

src/fmu/tools/rms/set_data_constant.py
JiZur: We use lines like: `grid = project.grid_models[gridname]` where project is recognized as RMS project structure only if roxar library is on.

JiZur:
> Documentation is missing? I see it was in some of the earlier commits, but it has been removed.
documentation available here:
https://github.com/equinor/fmu-tools/blob/qcreset_set_data_constant/docs/qcreset.rst

src/fmu/tools/rms/set_data_constant.py
HansKallekleiv: Would it work to put it into a try/except clause to make the linting happy?
e.g.:
```
try:
    import roxar
except ImportError:
    pass
```

HansKallekleiv (approved):

PR #93, opened 2020-12-16
Add example json volumetrics

berland:
The table in this file can be parsed into a DataFrame like this::

```
from pathlib import Path
import json
import pandas as pd
js = json.loads(Path('1_geogrid_vol_oil_1.json').read_text())
df = pd.DataFrame.from_records(js["table"]["data"], columns=js["table"]["columns"])
```


jcrivenaes:
What is the use case?

berland:
> What is the use case?

This is a potentially way-too-late suggestion for Emerson on how to export RMS  volumetrics and other data, as both txt files and xlsx export of different RMS-data requires writing dedicated parsers.

mferrera:
Closing this PR as just a bit stale :-) (Julevask)

Issue #88, opened 2020-12-11
XLRDError from most recent xlrd


berland:
Crash in github actions, due to updated upstream library.

```
>       snorrebergdesign = summarize_design(
            testdir + "/data/sensitivities/distributions/" + "design.xlsx", "DesignSheet01"
        )

tests/test_calc_tornado.py:27: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/site-packages/fmu/tools/sensitivities/_designsummary.py:56: in summarize_design
    dgn = pd.read_excel(filename, sheetname)
/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/site-packages/pandas/util/_decorators.py:296: in wrapper
    return func(*args, **kwargs)
/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/site-packages/pandas/io/excel/_base.py:304: in read_excel
    io = ExcelFile(io, engine=engine)
/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/site-packages/pandas/io/excel/_base.py:867: in __init__
    self._reader = self._engines[engine](self._io)
/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/site-packages/pandas/io/excel/_xlrd.py:22: in __init__
    super().__init__(filepath_or_buffer)
/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/site-packages/pandas/io/excel/_base.py:353: in __init__
    self.book = self.load_workbook(filepath_or_buffer)
/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/site-packages/pandas/io/excel/_xlrd.py:37: in load_workbook
    return open_workbook(filepath_or_buffer)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

filename = '/home/runner/work/fmu-tools/fmu-tools/tests/data/sensitivities/distributions/design.xlsx'
logfile = <_io.TextIOWrapper name="<_io.FileIO name=6 mode='rb+' closefd=True>" mode='r+' encoding='utf-8'>
[...]

PR #94, opened 2020-12-21
Read xlsx

tralsos:
Changes pd.read_excel to use engine openpyxl since newer versions of xlrd does not support '.xlsx' files. Closes issue #88 . Requirements are updated to use pandas>1.1 and xlrd>2.0, but shold also work with previous requirements if we don't want to change these yet.


tralsos:
@jcrivenaes should we change the requirements as suggested here, or are there reasons to keep the previous requirements ?

berland:
We should not enforce versions that are not yet in komodo stable. Pandas 1.1 is on its way in now, but xlrd might take another month or so.

src/fmu/tools/sensitivities/_designsummary.py
```diff
@@ -53,15 +53,17 @@ def summarize_design(filename, sheetname="DesignSheet01"):
...
         dgn = pd.read_csv(filename)
     else:
         raise ValueError(
-            "Design matrix filename should be on Excel or csv format"
-            " and end with .xlsx or .csv"
+            "Design matrix should be on Excel or csv format"
+            " and filename should end with .xlsx or .csv"
         )
-
+    print(dgn.loc[0]["SENSCASE"])
```
berland, line 66: Is this a debugging statement?

src/fmu/tools/sensitivities/_designsummary.py
```diff
@@ -53,15 +53,17 @@ def summarize_design(filename, sheetname="DesignSheet01"):
...
+        dgn.dropna(axis=0, how="all", inplace=True)
+        dgn = dgn.loc[:, ~dgn.columns.str.contains("^Unnamed")]
     elif filename.endswith(".csv"):
         dgn = pd.read_csv(filename)
     else:
         raise ValueError(
-            "Design matrix filename should be on Excel or csv format"
-            " and end with .xlsx or .csv"
+            "Design matrix should be on Excel or csv format"
+            " and filename should end with .xlsx or .csv"
```
berland, line 64: "should -> must" is perhaps more to the point (.XLSX would even fail here, but we can live with that.)

src/fmu/tools/sensitivities/_designsummary.py
```diff
@@ -53,15 +53,17 @@ def summarize_design(filename, sheetname="DesignSheet01"):
 
     # Read design matrix and find realisation numbers for each sensitivity
     if filename.endswith(".xlsx"):
-        dgn = pd.read_excel(filename, sheetname)
+        dgn = pd.read_excel(filename, sheetname, engine="openpyxl")
+        dgn.dropna(axis=0, how="all", inplace=True)
```
berland, line 57: Is this due to things like "coloured cells"? If so, maybe nice to add a comment about that (to explain why the line of code is there, otherwise it could disappear in later refactorings)

tralsos: Have added this

berland:
Test pass on combinations of xlrd 1.2.0/2.0.1, and pandas versions 0.25 up to 1.2.0. Except Pandas 1.1.1, but this particular version is not in use.

berland (approved):

PR #96, opened 2021-01-08
Sort test data into directories pr. tool

### LINKED ISSUE #87: Directory layout in tests directory
fmu-tools is growing, and thus the test-directory will get chaotic unless action is taken.

Suggestion is to replicate the directory structure in `src/fmu/tools` under `tests`, so that when there is a directory `src/fmu/tools/rms`, there is also a `tests/rms` in which   tests for rms tooling is placed. Data directories is placed alongside `test_*.py` files, unless the data is to be used by more test code - then it may be placed higher up in the test data structure.

This layout resembles/inspired by https://github.com/equinor/semeio which is a similar package with a collection of somewhat independent tools.

berland:
Resolves #87 

Bonus: Replace os.path by pathlib.


tralsos (approved):
This looks cleaner. Thanks!

PR #97, opened 2021-01-08
Pylint adjustments for qcforward

berland:
Solves only the lowest hanging pylint fruits.

Remove unwarranted pylint exception.


jcrivenaes (approved):
Looks good; seems it take some time to get rid of 2.7 habits....

jcrivenaes (approved):

PR #103, opened 2021-01-15
Add pre-commit.

berland:
Developers can do 'pre-commit install' to install in local clone (after pip install pre-commit)


HansKallekleiv (approved):
nice 👍

PR #105, opened 2021-01-26
bug fix for properties without code names

tnatt:
XTGeo gave an empty dictionary as output for code names for properties that did not have any. This caused a bug when trying to replace codes in dicrete parameters with code names. 

As a bug fix the code values are used as code names for relevant properties.
(Note, the issue has been fixed in XTGeo now)

Also added support for using values as input in filter list

### DIFF
--- a/src/fmu/tools/qcproperties/_propstat.py
+++ b/src/fmu/tools/qcproperties/_propstat.py
@@ -202,6 +202,13 @@ class PropStat:
                         f"{param} parameter is not!"
                     )
                 codes = xtg_prop.codes.copy()
+
+                # if no codenames, use codevalues as codenames
+                if not codes:
+                    codes = {
+                        value: str(value)
+                        for value in np.unique(xtg_prop.get_active_npvalues1d())
+                    }
             else:
                 codes = self._wells[0].get_logrecord(param).copy()
 
--- a/src/fmu/tools/qcproperties/_propstat_parameter_data.py
+++ b/src/fmu/tools/qcproperties/_propstat_parameter_data.py
@@ -69,9 +69,10 @@ class PropStatParameterData:
         self._properties, self._selectors = self._input_conversion(
             properties, selectors
         )
-
         # combine data and set different instance attributes
         self._combine_data(filters)
+
+        self._filter_values_to_string()
 
     @property
     def properties(self):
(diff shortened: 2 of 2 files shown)


src/fmu/tools/qcproperties/_propstat.py
```diff
@@ -202,6 +202,13 @@ def _codes_to_codenames(self, dframe):
                         f"{param} parameter is not!"
                     )
                 codes = xtg_prop.codes.copy()
+
+                # if no codenames, use codevalues as codenames
+                if not codes:
+                    codes = {
+                        value: str(value)
+                        for value in np.unique(xtg_prop.get_active_npvalues1d())
+                    }
```
berland, line 211: In case you don't want to inject "fake" dictionary values, this could perhaps alternatively be solved using the `<dict>.get()` function. If you have a dict `somedict`, you can do `somedict.get("somekey", "somekey")` which will return "somekey" if "somekey" is not in the dictionary:
```
In [1]: a = dict(foo="bar")

In [2]: a
Out[2]: {'foo': 'bar'}

In [3]: a.get("foo")
Out[3]: 'bar'

In [4]: a.get("com", "com")
Out[4]: 'com'
```

jcrivenaes: I don't think there is a risk for fake dict values in tnatt's approach(?)

tnatt: I am not sure what you mean by "fake" values @berland ? All of the values inserted should exist in the property.

berland: Fake is probably an incorrect term here then ;)  Call then "undefined". The question was more if you really want to provide a modified dictionary for the default values, or if you want to code the defaulting into the code that accesses the dictionary.

berland:
Test code?

jcrivenaes:
Tests should be in xtgeo, but a bit tricky to make it fluent on GH actions as this is a Roxar API related. Perhaps mocking is the solution.

tnatt:
Sorry for the late reply... I will add a test code for using integers as input in the filters, and can talk to you @jcrivenaes regarding mocking. Would be nice to see how that works :)
From next week I am available and will dive into fmu.tools 👍  I have several tasks on my to-do list and will revisit this PR then.

jcrivenaes:
I would propose to skip this as it is ficed in latest xtgeo

jcrivenaes:
Should we do something here og just close?

tnatt:
We skip this one and close it.

PR #106, opened 2021-01-28
Remove docs on deprecated sensitivity code

berland:
I think we better can remove this docs, as the tornadoplot is deprecated, and then   I don't think one needs the `summarize_design` example?


tralsos:
This part of the repository is in use. Currently we only have support in webviz for tornado plots for volumes and time series. If we want tornado plots for other repsonses we can use this to generate the tornado tables and then plot it in software of choice e.g. in jupyter notebooks.  summarize_design is also in use.  The code and documentation related to the old tornado plots in webportal has already been removed.

berland:
ok!

PR #107, opened 2021-01-28
Add tests for background values in fmudesign

berland:
Also give error messages in some faulty scenarios


tests/sensitivities/test_use_cases.py
```diff
@@ -0,0 +1,90 @@
...
+            "sensname",
+            "numreal",
+            "type",
+            "param_name",
+            "dist_name",
+            "dist_param1",
+            "dist_param2",
+        ],
+        data=[
+            ["ref", None, "ref", None],
```
tralsos, line 62: I wonder if the sensitivity type here should be "background" instead of "ref". (The sensname can still be "ref" of course). The difference is that the intention with a "ref" sensitivity is that it is normally only one realisation, and would not be plotted in the tornado plot, but only used as a reference to calculate the deltas in the plot. The background sensitivity will pop up as a sensitivity in the plot and you can see the P10/P90 variation in the background realisations. This might be useful to compare the variation of the scenario sensitivities to the variation in the background (ref) realisations. But I leave this to you and your use case here. The test looks fine.

tralsos (commented):
This looks good, thanks. 
To comment on the default values for background: The reason why also the background parameters require default values is that they are allowed to be set up similarly as in the designinput, and have several scenarios. Also if for a single sensitivity the number of realisations are set larger than the number of realisations in the background, this results in a warning and the remaining realisations are filled with background default values.

tests/sensitivities/test_use_cases.py
berland: changing the second `"ref"` to `"background"`

tralsos (approved):
Approved

PR #115, opened 2021-02-12
Support alternative traj name in create_rft_ertobs

berland:


src/fmu/tools/rms/create_rft_ertobs.py
```diff
@@ -41,23 +41,24 @@
...
-    "absolute_error",
+    "loglevel",
+    "project",
     "relative_error",
+    "rft_prefix",
     "rms_name",
-    "ecl_name",
-    "zonename",
-    "gridname",
+    "trajectory_name",
```
berland, line 58: This is the only change in this list, the rest is just alphabetical sorting.

tests/rms/test_create_rft_ertobs.py
```diff
@@ -157,6 +162,19 @@ def test_get_well_coords():
         == np.array([[0, 0, 0, 0], [2000, 0, 0, 2000]])
     ).all()
 
+    # Alternative trajectory name in RMS:
+    assert (
+        create_rft_ertobs.get_well_coords(
+            rms_project_mock, "R-99", trajectory_name="Imported trajectory"
+        )
+        == np.array([[0, 0, 0, 0], [2000, 0, 0, 2000]])
+    ).all()
```
JiZur, line 171: Should it not be an error message sent in case assert test is False ?

berland: Did not understand? This is not an error condition, so no error message.

JiZur (approved):
Seems alright :)

PR #118, opened 2021-03-02
Test multiple Pandas version.

berland:
Some RMS versions are still using pandas < 1.0

### DIFF
--- a/.github/workflows/fmu-tools.yml
+++ b/.github/workflows/fmu-tools.yml
@@ -22,6 +22,12 @@ jobs:
     strategy:
       matrix:
         python-version: ["3.6", "3.7", "3.8"]
+        pandas-version: ["0.21.0", "1.*"]
+        exclude:
+          - python-version: "3.8"
+            pandas-version: "0.21.0"
+          - python-version: "3.7"
+            pandas-version: "0.21.0"
 
     steps:
       - name: Checkout commit locally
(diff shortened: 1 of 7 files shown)


tralsos:
Could you please add what the issue is, and what these changes are doing ?

berland:
> Could you please add what the issue is, and what these changes are doing ?

Sorry.. See #122

jcrivenaes (approved):
We could simplify:
```
@pytest.mark.skipif(
    version.parse(pd.__version__) < version.parse("0.25.0"),
    reason="Pandas 0.25.0 is required for fmudesign",
)
```
via `conftest.py`, but can take that later

jcrivenaes:
WRT to pandas from fmu-design, I will think in practice it will work since you use it in Komodo env, not in RMS.

If you want to be on the very safe side you could add a version test in you script:

```
from packaging import version
import pandas as pd
if version.parse(pd.__version__) < version.parse("0.25.0"):
    raise RuntimeError("Your pandas version is too old, requires version >= 0.25")
```

berland:
Yes, this would only be a problem for those that run their own virtual environment, and for some reason wants both fmu_design and an old Pandas version. My guess is that those people count up to zero.

In general, it is too much work to specify minimum versions of all imported libraries in every tool, and it would not be considered reliable.  Komodo solves it by specifiying a specific release for every package, and explicitly tests the chosen combination.

PR #120, opened 2021-03-03
Make create_rft_ertobs support time-variation in observations

berland:
Different observations at different well-points at different dates
require a peculiar setup that must be compatible with GENDATA_RFT in semeio
and GENDATA in Ert.


berland:
The code in semeio that proves this setup has not been merged yet, is in https://github.com/equinor/semeio/pull/304

jcrivenaes (approved):
LGTM

PR #121, opened 2021-03-05
Add set_data_empty method to qc reset and edit init file

JiZur:


src/fmu/tools/rms/qcreset.py
```diff
@@ -209,3 +208,194 @@ def set_surfaces(
...
+            of the grid models where properties need to be made empty. The
+            value associated to this key should be a list of property names to
+            modify. If a string ``all`` is assigned instead of a list, all
+            available properties for this grid model name will be used.
+            Alternatively, if a list of grid models names is given instead of a
+            dictionary, the method will apply to all properties within these
+            grid models.
+    """
+
+    assert isinstance(config, dict), "Argument must be a Python dictionary!"
```
berland, line 259: This would raise the exception `AssertionError`, while I believe a `TypeError` would be more correct.

src/fmu/tools/rms/qcreset.py
```diff
@@ -209,3 +208,194 @@ def set_surfaces(
...
+                print(" >> " + data_type)
+                surf_names = dict_val[data_type]
+                if isinstance(surf_names, str):
+                    if surf_names == "all":
+                        for surface in surfaces:
+                            set_safe_empty(
+                                project, surf_type, surface.name, data_type
+                            )
+                    else:
+                        raise Exception(
```
berland, line 327: `ValueError` better?

src/fmu/tools/rms/qcreset.py
```diff
@@ -209,3 +208,194 @@ def set_surfaces(
...
+        if isinstance(config["grid_models"], list):
+            # work directly at grid models level
+            for gridname in config["grid_models"]:
+                print(" >> " + gridname)
+                grid = project.grid_models[gridname]
+                for prop in grid.properties:
+                    try:
+                        prop.set_empty()
+                        print(" >> >> " + prop.name)
+                    except Exception as e:
```
berland, line 364: Should not catch every thinkable exception.

berland:
Possible to add some tests?

JiZur:
Problem is that it is meant to work only in RMS

iainashepherd:
Is testing needed for this? If so, can Jan's new approach for testing within RMS help the situation?

src/fmu/tools/rms/qcreset.py
```diff
@@ -209,3 +209,179 @@ def set_surfaces(
...
+            dictionary, the method will apply to all properties within these
+            grid models.
+    """
+
+    if not isinstance(config, dict):
+        raise TypeError("Argument must be a Python dictionary!")
+    assert "project" in config.keys(), "Input dict must contain key 'project'!"
+    project = config["project"]
+    assert "value" in config.keys(), "Input dict must contain key 'value'!"
+    value = config["value"]
```
HansKallekleiv, line 265: This appears not to be used for anything?

src/fmu/tools/rms/qcreset.py
```diff
@@ -209,3 +209,179 @@ def set_surfaces(
...
+
+    def set_safe_empty(
```
HansKallekleiv, line 267: Move this function outside the `set_data_empty` function?

src/fmu/tools/rms/qcreset.py
```diff
@@ -209,3 +209,179 @@ def set_surfaces(
...
+            else:
+                surf = project.zones[name][data_type]
+            surf.set_empty()
+            print(" >> >> " + name)
+        except Exception as e:
+            print(" >> >> " + name + " cannot be modified")
+            print(e)
+
+    def set_surfaces(
+        project: _roxar.Project, surf_type: str, dict_val: Union[List, Dict]
```
HansKallekleiv, line 290: Move this function outside the `set_data_empty` function?

HansKallekleiv (changes requested):

src/fmu/tools/rms/qcreset.py
JiZur: We could do that indeed, but do we want users to be able to use those functions independently ?

src/fmu/tools/rms/qcreset.py
JiZur: it's actually used as argument in the sub-routines, isn'it ?

JiZur: my bad, value is obviously not used in set_data_empty, will remove it

JiZur:
@HansKallekleiv : it now complains when building doc that _roxar is not defined in the last tests

src/fmu/tools/rms/qcreset.py
```diff
@@ -0,0 +1,383 @@
...
+            raise TypeError(
+                "Value associated with key 'zones' must be of type list or dict!"
+            )
+
+    print("End of function set_data_constant().")
+
+
+def set_data_empty(config: Dict):
+    """Set data from RMS empty.
+
```
jcrivenaes, line 287: Why  use input `config: Dict? ` Should it not only use keywords as given in the docstring?

HansKallekleiv (approved):

PR #126, opened 2021-03-26
Fix doc and xx print strings

jcrivenaes:
A couple of issues when running actual testing in RMS:
* Documentation has errors
* Get XX prints that should be removed


berland (approved):
🚢

Issue #127, opened 2021-04-06
Updates for examples used in fmu-tools documentation


iainashepherd:
**qcforward functions**

1) Compare well zonation vs well (Example when ran inside RMS with different settings for wells)

- "gridzones": [ZONEGRIDNAME]      believe this should be        "gridprops": [ZONEGRIDNAME]
- mode: "write"    /    mode: "append"      should be     "mode": "write"    /   "mode": "append"
- Script fails if qc folder doesn't exist or file named "well_vs_grid.csv" doesn't exist
- Not correctly reporting SET1 or SET2 stats to file.

2) Grid Quality Indicators (Example when ran inside RMS)

- qcf.run(usedata, project=project)     believe this should be     QCJOB.run(usedata, project=project)
- mode: "write"     should be    "mode": "write"  
- Script fails if qc folder doesn't exist or file named "gridquality.csv" doesn't exist

3) Running grid_statistics  (Example when executed inside RMS (More settings)

- "path": PATH,   should be     "path": "PATH",

**The qcproperties class**

1) Examples scripts
- Scripts do not compute anything in RMS when completed - should these not at least say "Done"?

**The qcreset module**

1) set_data_constant examples (something similar probably needed for set_data_empty examples?)
- from fmu.tools import qcreset     should be    from fmu.tools.rms import set_data_constant as qcreset

JiZur:
The last point will be corrected with the PR as the file containing the code is renamed into qcreset. So the import: "from fmu.tools.rms import qcreset" should work fine

Issue #112, opened 2021-02-11
adding the number of points in the statisitcs 


miroine:
is it possible to include the sample size when we extract the statistics in `QCProperties` class, this is quite important to know how reliable are the statistical numbers

tnatt:
Thanks for the suggestion @miroine, that should be a simple fix.

Issue #111, opened 2021-02-11
getting fraction per zone as a statistics for discrete parameter


miroine:
fraction per zone for discrete parameter is very useful (example: sand percentage for each zone), is it possible to include that in the `QCProperties `class (or is it there and I missed it ?) 
`xtgeo` has quite nice well class called `._get_fraction_per_zone()`   that I have used before ? how can we incorporate that in the qcproperties ?

tnatt:
Hi @miroine, that should be possible with existing functionality. Just add the discrete Facies/NTG parameter as a property and add the Zone parameter as a selector. The trick is that you need to add a disc=True argument for the method that writes out the csv. 

This functionality was added a bit later, and I realize not documented properly... will make a note of that :) 
Please test and see if it works as you expect! Thanks

![image](https://user-images.githubusercontent.com/61694854/107647154-1f7d1380-6c7b-11eb-8d70-5c7ebe10d9ab.png)

tnatt:
To improve user-friendliness I will change the code to make the disc=True argument redundant when properties of the same type (discrete or continuous) are used as input 👍

PR #128, opened 2021-04-13
Refactor QCProperties

### LINKED ISSUE #123: Documentation: qcproperties
Some comments on the documentation for qcproperties

1. The general name for the functions in qcforward is check() – which makes sens since one is checking stuff. For qcproperties I find it strange that 5 out of 8 examples got check() while the remaining 3 got extract_statistics() which is a better name
2.	One uses the names Geo**G**rid and Sim**G**rid in all examples except one where Geogrid and Simgrid is used. (“Comparison of data from different sources”)
3.	Wildcards for well name is great. But there is not much explanation. I would like to have a sentence explaining that a dot a star and a dollar (.\*$) means “the rest of the string can be anything”. I guess the dot is important for telling the system that wildcards are coming? Are there other wildcards to use too ?  (^*.)  or (.?.) or can general regex-syntax be used?
4.	The final setup with everything defined in a single yml-file is good. But to better see the structure (for those not that familiar with python dict vs yml) it could be useful if the yaml-file did exactly the same (or more) as the example above (“Example below collects statistical data from four different sources” )

tnatt:
Refactoring of QCProperties which includes: 
- Improved readability of code
- Removal of old/unnecessary functionality
- Cleanup of tests
- Documentation updated with latest changes and user feedbacks #123 #127 

API changes
- Only properties of the same type (continous/discrete) allowed in a run. This avoids the confusion of having two dataframes whitin a QCProperties instance ref #111 
- No longer allowed to specify "csvfile" in the input dictionary
- Statistics for discrete properties is now reported in fractions instead of percentages. Column names for discrete properties changed from "Percentage" and "Percentage_Weighted" to "Avg" and "Avg_Weighted" in alignment with the column names used for continous statistics. 
- A Pandas dataframe is returned when running an instance method for extracting statistics instead of a PropStat instance

Added functionality 
- A "range" option is added to filters to enable using continous properties as filters. It is also possible to add "range" as a value directly on a property to remove unwanted high/low values from the statistics.
- "COUNT" added as a clolumn to continous statistics to be able to see sample size #112 

This PR closes #123 #112 #127 #111


HansKallekleiv (approved):
Looks good! Like the use of test classes 👍

PR #133, opened 2021-05-14
Move fipmapper from subscript to fmu-tools

berland:


src/fmu/tools/fipmapper/fipmapper.py
```diff
@@ -0,0 +1,320 @@
...
+        assert isinstance(self._mapdata, dict), "FipMapper needs a dictionary"
+
+        # Determine our capabilities:
+        self.has_fip2region = "fipnum2region" in self._mapdata
+        self.has_fip2zone = "fipnum2zone" in self._mapdata
+        self.has_region2fip = "region2fipnum" in self._mapdata
+        self.has_zone2fip = "zone2fipnum" in self._mapdata
+
+    def _get_explicit_mapdata(self, yamldata: Dict[str, Any]):
+        """Fetch explicit mapping configuration from a dictionary,
```
jcrivenaes, line 88: End first sentence with period ., not comma ,

src/fmu/tools/fipmapper/fipmapper.py
```diff
@@ -0,0 +1,320 @@
...
+
+
+class FipMapper:
+    def __init__(
+        self,
+        yamlfile: Union[str, Path] = None,
+        mapdata: Dict[str, str] = None,
+        skipstring: Union[List[str], str] = None,
+    ):
+        """FipMapper is a utility class for being able to map between
```
jcrivenaes, line 27: Google docstrings: First line shall be a one liner summary.

https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

jcrivenaes (commented):
Is there documentation on this?

berland:
> Is there documentation on this?

Only what you see in the source.

src/fmu/tools/fipmapper/fipmapper.py
berland: Fixed.

src/fmu/tools/fipmapper/fipmapper.py
berland: Fixed.

jcrivenaes:
> Is there documentation on this?
> Only what you see in the source. The user facing part of it is documented in prtvol2csv in subscript.

Would prefer to have some documentation here also, eventually that it points to the documentation in subscript

berland:
> > Is there documentation on this?
> > Only what you see in the source. The user facing part of it is documented in prtvol2csv in subscript.
> 
> Would prefer to have some documentation here also, eventually that it points to the documentation in subscript

As RST docs? The FipMapper class is in the current state not meant for users, only for applications (pr now only prtvol2csv). As an API, inline doc should be sufficient for FipMapper?

jcrivenaes:
OK, if it is not meant for users it is ok

jcrivenaes (approved):

Issue #130, opened 2021-04-27
[rmsvolumetrics] Change name from PORE_ to PORV_


berland:
https://github.com/equinor/fmu-tools/blob/70232980fa51458b08245bb2a7f92dd595db8469/src/fmu/tools/rms/volumetrics.py#L81

Suggested by @rnyb to change from `PORE` to `PORV`.

This could make sense, but would be an API break with potential consequenses.

Which other tools is it that this code should align to?

berland:
https://wiki.equinor.com/wiki/index.php/FMU_standards/Volumetrics

PR #136, opened 2021-05-26
Change column name PORE_x to PORV_x in volumetrics

berland:
This is a bugfix, the previous column name was not according
to the defined standard.

Solves #130


HansKallekleiv (approved):
👍

PR #137, opened 2021-05-27
Add type hinting to some modules, enforce in CI

berland:
It is up for discussion whether we want to enforce that mypy passes through, it will create some extra hurdles for contributors.


jondequinor:
Yes. Still, changes to API that is already type hinted, should be required to uphold the level of type hinting.

jcrivenaes (approved):
Yes it is some work to implement this, but it will give better code in long run, and type hinting also improves docstring generation. So it LGTM.

PR #142, opened 2021-06-17
Support integer inputs to fipmapper for reg/zone

berland:
The return type of disjoint_sets() is always
strings for regions and zones, but other functions in
fipmapper try to conserve the datatype obtained from
the yaml parsing.

Keeping integers as ints is done in order to facilitate
sorting of regions and zones.

### DIFF
--- a/src/fmu/tools/fipmapper/fipmapper.py
+++ b/src/fmu/tools/fipmapper/fipmapper.py
@@ -175,17 +175,26 @@ class FipMapper:
     def get_regions(self) -> List[str]:
         """Obtain a sorted list of the regions that exist in the map"""
         assert "region2fipnum" in self._mapdata, "No data provided for regions"
-        return sorted(self._mapdata["region2fipnum"].keys())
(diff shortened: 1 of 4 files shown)


src/fmu/tools/fipmapper/fipmapper.py
```diff
@@ -218,34 +227,51 @@ def fip2region(self, fip: Optional[int]) -> List[str]:
...
         try:
             fips = self._mapdata["region2fipnum"][region]
             if not isinstance(fips, list):
                 # Single FIPNUM in input
                 return [int(fips)]
             return [int(fip) for fip in fips]
         except KeyError:
+            import pdb
+
+            pdb.set_trace()
```
berland, line 257: ouch

berland:
This is now confirmed to work on a big dataset. Accompanying script might appear in subscript.

Ready to be reviewed.

jcrivenaes (approved):
Looks OK

PR #156, opened 2021-08-25
Change SET column from fipmapper when FIPNUM is one-to-one with SET

### LINKED ISSUE #155: [fipmapper] Align SET enumeration with FIPNUM when possible
It is typical for the FIPNUMs to be one-to-one to the sets, and this might also become a recommended practice. 

It is suggested that the disjoint-sets code in fipmapper detects this situation, and re-enumerates the sets so that it matches the FIPNUMs exactly.

berland:
Resolves #155


berland:
This has been solved in Webviz

PR #158, opened 2021-10-01
Qcforward blockedwells vs property

jcrivenaes:
Add method to detect equalness between grid properties and blocked wells


docs/qcforward/bw_vs_gridprops.inc
```diff
@@ -0,0 +1,183 @@
...
+  A string to identify the data set. Recommended.
+
+show_data
+  A string (or bool) if the full dataset shall be shown to screen as a dataframe
+  (table). This can be useful when investigating what wells that triggers e.g. a stop.
+  Possible values are False (default, means no output), "warn" to wells with WARN or
+  STOP criteria, "fail" to see wells that fails only (due to STOP) and True or "all" to
+  show everything. Any other words (like "foo") will be interpreted as "all").
+
+
```
tnatt, line 53: Add documentation of `tolerance `argument

src/fmu/tools/qcforward/_blockedwells_vs_gridprops.py
```diff
@@ -0,0 +1,328 @@
...
+        "nametag": "MYDATA1",
+        "verbosity": "debug",
+        "path": PATH,
+        "bwells": WELLFILES,
+        "grid": GRIDFILE,
+        "gridprops": [["PHIT", ROFFFILE], ["FACIES", ROFFFILE]],
+        "compare": {"Facies": "FACIES", "PHIT": "PHIT"},  # bwname: modelname
+        "actions": [
+            {"warn": "anywell < 80%", "stop": "anywell < 70%"},
+            {"warn": "allwells < 90%", "stop": "allwells < 80%"},
```
tnatt, line 35: `anywell` and `allwells` is this an old syntax? I only see `any` and `all` in the documentation.

src/fmu/tools/qcforward/_blockedwells_vs_gridprops.py
```diff
@@ -0,0 +1,328 @@
...
+        ],
```
tnatt, line 36: Same comment as above

src/fmu/tools/qcforward/_blockedwells_vs_gridprops.py
```diff
@@ -0,0 +1,328 @@
...
+If one need to have different actions for different comparisons, use e.g.:
+
+::
+
+    DATA2 = deepcopy(DATA1)
+    DATA2["compare"] = {"VSH": "Vshale"}
+    DATA2["actions"] =
+        [
+            {"warn": "anywell < 86%", "stop": "anywell < 75%"},
+            {"warn": "allwells < 97%", "stop": "allwells < 89%"},
```
tnatt, line 51: Same comment as above

docs/qcforward/bw_vs_gridprops.inc
```diff
@@ -0,0 +1,183 @@
...
+nametag
+  A string to identify the data set. Recommended.
+
+show_data
+  A string (or bool) if the full dataset shall be shown to screen as a dataframe
+  (table). This can be useful when investigating what wells that triggers e.g. a stop.
+  Possible values are False (default, means no output), "warn" to wells with WARN or
+  STOP criteria, "fail" to see wells that fails only (due to STOP) and True or "all" to
+  show everything. Any other words (like "foo") will be interpreted as "all").
```
tnatt, line 51: Would it be better to have this as an optional argument with default value None? That would reduce the list of possible values to `warn`/`fail`/`all` and make the documentation a bit simpler.

src/fmu/tools/qcforward/_blockedwells_vs_gridprops.py
```diff
@@ -0,0 +1,328 @@
...
+                    wells.remove("all")
+                if wells:
+                    print(f"Wells with WARN or STOP are these {wells}:")
+                    print(comb[comb["WELLNAME"].isin(wells)].to_string())
+                else:
+                    print("No wells reported WARN or STOP")
+
+            else:
+                print(f"Wells with WARN or STOP or OK are:")
+                print(comb.to_string())
```
tnatt, line 180: This could probably be shortened a bit. Something like:

```
        if self.ldata.show_data:
            if self.ldata.show_data in ["fail", "warn"]:
		mask = (dfr["STATUS"] == "STOP") if self.ldata.show_data == "fail" else (dfr["STATUS"] != "OK")
                wells = [well for well in dfr[mask]["WELL"].unique() if well != "all"]
		if wells:
                    print(f"Wells within {self.ldata.show_data} criteria are these {wells}:")
                    print(comb[comb["WELLNAME"].isin(wells)].to_string())
                else:
                    print("No wells within {self.ldata.show_data} criteria")
            else:
                print(f"Wells with WARN or STOP or OK are:")
                print(comb.to_string())
```

src/fmu/tools/qcforward/_blockedwells_vs_gridprops.py
```diff
@@ -0,0 +1,328 @@
...
+        for bwprop, modelprop in self._ldata.compare.items():
+            usebwprop = bwprop
+            usemodelprop = modelprop
+            if bwprop == modelprop:
+                usebwprop = bwprop + "__bw"
+                usemodelprop = modelprop + "__model"
+            dname = bwprop + ":" + modelprop
+            dnameflag = dname + "_flag"
+            comb = self._eval_tolerance(comb, usebwprop, usemodelprop, dname, dnameflag)
+            diffs[dname] = dnameflag
```
tnatt, line 225: Maybe skip some unnecessary variables here?

```
        for bwprop, modelprop in self._ldata.compare.items():
            dname = bwprop + ":" + modelprop
            dnameflag = dname + "_flag"
            if bwprop == modelprop:
                bwprop = bwprop  + "__bw"
                modelprop = modelprop + "__model"
            comb = self._eval_tolerance(comb, bwprop, modelprop, dname, dnameflag)
            diffs[dname] = dnameflag
```

src/fmu/tools/qcforward/_blockedwells_vs_gridprops.py
```diff
@@ -0,0 +1,328 @@
...
+        comb = df_in.copy()
+        tol = self.ldata.tolerance
+
+        relative = False
+        if isinstance(tol, float):
+            tolerance = tol
+        elif isinstance(tol, dict):
+            if "rel" in tol.keys():
+                relative = True
+            tolerance = list(tol.values())[0]
```
tnatt, line 240: Might be better to have relative as an oneliner
`relative = isinstance(tol, dict) and "rel" in tol`

This could also be done for the tolerance, but less readable maybe 🙂 
`tolerance = tol if isinstance(tol, float) else list(tol.values)[0]`

src/fmu/tools/qcforward/_qcforward.py
```diff
@@ -135,14 +145,17 @@ def evaluate_qcreport(self, dfr, name):
                 dfr_status_print = dfr_status.to_string()
                 print(f"{dfr_status_print}\n", file=stream)
                 if status == "STOP":
-                    QCC.force_stop("STOP criteria is found!")
-
+                    if stopaction:
+                        QCC.force_stop("STOP criteria is found!")
+                    else:
+                        return status
```
tnatt, line 151: Unnecessary else statement

```
if status == "STOP": 
    if stopaction:
        QCC.force_stop("STOP criteria is found")
    return status
```

docs/qcforward/bw_vs_gridprops.inc
```diff
@@ -0,0 +1,183 @@
...
+  filename, for example ``[["Zone", "zone.roff"]]``. This is required when running
+  outside RMS but not needed when run inside RMS.
+
+
+Example when run inside RMS
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. code-block:: python
+
+    import pathlib
```
tnatt, line 87: Change to `from pathlib import Path`
To avoid NameError saying 'Path' is not defined

src/fmu/tools/qcforward/_blockedwells_vs_gridprops.py
tnatt: After testing on Gullfaks with many wells, I realize that if the user select `fail `or `warn `as the desired output, we should filter away rows that is flagged as "MATCH" to highlight the inconsistent data!

src/fmu/tools/qcforward/_qcforward.py
```diff
@@ -197,3 +210,47 @@ def parse(self, rule):
...
+                "Both criteria 'warn' and 'stop' are required in actions! "
+                f"You have keys: {list(elem.keys())}"
+            )
+        vals = list(elem.values())
+        if "any" in str(vals) and "all" in str(vals):
+            # avoid that one entry mixes all and any
+            raise ValueError(
+                "Seems wrong to mix 'all' and 'any' in one line; be consistent"
+            )
+
```
tnatt, line 255: Should also include an Error if the action does not contain `all` or `any` to avoid error later

tnatt (changes requested):
Overall I like the functionality, it is quite fast to use and offers a quick qc to secure that your parameters have been conditioned to 👍 🙂 

For facies this can be a good tool if the user have created a facies parameter by merging different outputs from different facies modelling jobs, this can easily lead to mismatch between the BW log and the grid parameter.

I think it might be useful to include some filter options though. e.g, when looking at Sw match it would be beneficial to filter away the water zone. Also for other parameters it would be nice being able to ignore parts of the BW logs if it is not of interest.

Mainly the code looks ok, I have made some comments. My main concern from a user perspective is the actions. At the moment there is not enough checks on the format of these and it is easy to make mistakes. For this qcforward test maybe we should change the format of the actions, the user will e.g. never want to look at a match above a certain percentage so we actually don't need the "<" sign. Maybe actions could be a dictionary then it is easier to check the input and we avoid having so many checks on a string? What do you think? 

```
actions = {
    "warn": {"any": 90, "all": 95}, 
    "stop": {"any": 80, "all": 90},
}
```

As a last comment it could be nice to add typehints, probably we would like to include mypy into the CI for fmu.tools at some point 🙂

src/fmu/tools/qcforward/_blockedwells_vs_gridprops.py
jcrivenaes: The code will look at substrings "all" or "any", meaning that both "all", "allwells", "allwhatever" will work equally. I was thinking that "allwells" in this case is more readable than "all"?

src/fmu/tools/qcforward/_blockedwells_vs_gridprops.py
jcrivenaes: Yes some regards ternary if (one-liners) as less readable

src/fmu/tools/qcforward/_qcforward.py
jcrivenaes: Possibly

src/fmu/tools/qcforward/_blockedwells_vs_gridprops.py
jcrivenaes: > Overall I like the functionality, it is quite fast to use and offers a quick qc to secure that your parameters have been conditioned to 👍 🙂
> 
> For facies this can be a good tool if the user have created a facies parameter by merging different outputs from different facies modelling jobs, this can easily lead to mismatch between the BW log and the grid parameter.
> 
> I think it might be useful to include some filter options though. e.g, when looking at Sw match it would be beneficial to filter away the water zone. Also for other parameters it would be nice being able to ignore parts of the BW logs if it is not of interest.

Making such filters can be a challenge if the field have many contact compartments, and/or individual wells are influenced by production. One possibly simpler option is to create a single filter value for a group of wells (but I need in case look into the details). A side note is that SW is usually not conditioned to blocked wells; the main purpose with this is blocked wells that are used in hard-conditioning (porosity, permeability, facies, ...) 

> 
> Mainly the code looks ok, I have made some comments. My main concern from a user perspective is the actions. At the moment there is not enough checks on the format of these and it is easy to make mistakes. For this qcforward test maybe we should change the format of the actions, the user will e.g. never want to look at a match above a certain percentage so we actually don't need the "<" sign. Maybe actions could be a dictionary then it is easier to check the input and we avoid having so many checks on a string? What do you think?

In some other qcforward jobs both > and < are applied. So perhaps consistency between jobs is good? I also think the signs add readability. Otherwise I agree that better tests on the actions is beneficial, so I will take a look

> 
> ```
> actions = {
>     "warn": {"any": 90, "all": 95}, 
>     "stop": {"any": 80, "all": 90},
> }
> ```
> 
> As a last comment it could be nice to add typehints, probably we would like to include mypy into the CI for fmu.tools at some point 🙂

Yup, adding type hints is a good idea (I already have this in xtgeo for all new code)

src/fmu/tools/qcforward/_blockedwells_vs_gridprops.py
```diff
@@ -154,29 +158,41 @@ def run(
...
+        # make it possible to print the underlying dataframe, either some wells (.e.g
+        # the failing) or all wells. If 'fail' it will only show those lines that
+        # contains FAIL
+        show = self.ldata.show_data
+        if show and isinstance(show, dict):
+            if (
+                not isinstance(show, dict)
+                and "lines" not in show
+                and "wellstatus" not in show
+            ):
```
tnatt, line 170:
```suggestion
        if show is not None:
            if (
                not isinstance(show, dict)
                or ("lines" not in show)
                or ("wellstatus" not in show)
            ):
```

This Error will never happen since you are checking that show is on dictionary form on the first line. Maybe this is a better solution?

docs/qcforward/bw_vs_gridprops.inc
```diff
@@ -123,7 +153,8 @@ Example when run inside RMS
             "verbosity": "info",
             "tolerance": {"rel": 0.1},
             "nametag": "BWCHECK",
-            "show_data": "warn",
+            "show_data": {"lines": "FAIL", wellstatus": "stop"},
```
tnatt, line 156:
```suggestion
            "show_data": {"lines": "FAIL", "wellstatus": "stop"},
```

src/fmu/tools/qcforward/_qcforward.py
```diff
@@ -252,5 +260,12 @@ def actions_validator(actionsin: dict) -> dict:
             raise ValueError(
                 "Seems wrong to mix 'all' and 'any' in one line; be consistent"
             )
+        if "any" in str(vals):
+            anyhit = True
+        if "all" in str(vals):
+            allhit = True
+
+    if not anyhit or not allhit:
+        raise ValueError("One of 'any' or 'all' is missing in the actions input")
```
tnatt, line 269:
```suggestion
        if  ("any" not in str(vals)) or ("all" not in str(vals)):
            raise ValueError("One of 'any' or 'all' is missing in the actions input")
```
This Error check should be within the loop, otherwise only the last action will be checked. 
Another suggestion, you could check for "any" and "all" directly here, and drop the 6 lines related to the `anyhit` and `allhit` variables

src/fmu/tools/qcforward/_blockedwells_vs_gridprops.py
berland: `if isinstance(show, dict) and show:` I think will do it

PR #166, opened 2021-11-30
Avoid `xtgeo` prerelease dependency

anders-kiaer:
Using prerelease lower bound in `fmu-tools` requirements apparently open up for downstream packages getting prereleases installed (even though they haven't asked for prereleases with `--pre` AND there are new official releases of the dependency satisfying the requirement 🤔).


berland (approved):
+1

PR #168, opened 2021-12-17
Avg props along well for ensembles

jcrivenaes:


rnyb (approved):
LGTM

src/fmu/tools/ensembles/ensemble_well_props.py
```diff
@@ -0,0 +1,616 @@
...
+            self.reals = self.config["ensemble"]["realizations"]["entries"]
+        else:
+            raise ValueError("For 'realizations': 'range' or 'entries' is missing")
+
+        self.itera = self.config["ensemble"]["iteration"]
+
+        self.wellfile = self.config["well"]["file"]
+        self.lognames = self.config["well"].get("lognames", "all")
+        self.mdlog = self.config["well"].get("mdlog", "MDepth")
+        self.mdranges = self.config["well"]["mdranges"]
```
tnatt, line 211: Add a validation of mdranges intervals to ensure no intervals are overlapping and hence impacting the total cumulative length

tnatt (commented):
Tested the functionality and made some comments 🙂
Other than that it would be nice to have the example as a separate yaml-file which can be dumped directly to file and in addition be included in the documentation.
Also the documentation should describe the different arguments and if they are optional or required.

src/fmu/tools/ensembles/ensemble_well_props.py
```diff
@@ -0,0 +1,616 @@
...
+                            "'interval' as criteria"
+                        )
+
+                    propcase.codes = self._validate_codes(codes)
+                    propcase.interval = self._validate_interval(interval)
+
+        self.report_keep_intermediate = rpt.get("keep_intermediate_logs", False)
+        self.report_show_in_terminal = rpt.get("show_in_terminal", True)
+
+        self.config = None
```
tnatt, line 261: Is this line really needed? It creates some mypy complaints since the config is only type hinted as a dict.

jcrivenaes: Don't think so, so I will remove

tnatt (approved):
Remove the two print statements otherwise LGTM 👍

Issue #169, opened 2022-02-26
Scheduled github jobs fails


jcrivenaes:
Current scheduled github jobs fails on mypy and/or testing, e.g.:

https://github.com/equinor/fmu-tools/actions/runs/1901408286

PR #170, opened 2022-02-26
Fix recurrent testing errors

jcrivenaes:
Solving #169

### DIFF
--- a/src/fmu/tools/sensitivities/design_distributions.py
+++ b/src/fmu/tools/sensitivities/design_distributions.py
@@ -630,7 +630,7 @@ def _nearest_positive_definite(a_mat):
     kiter = 1
     while not _is_positive_definite(a3_mat):
         mineig = numpy.min(numpy.real(la.eigvals(a3_mat)))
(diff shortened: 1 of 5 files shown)


src/fmu/tools/sensitivities/design_distributions.py
```diff
@@ -630,7 +630,7 @@ def _nearest_positive_definite(a_mat):
     kiter = 1
     while not _is_positive_definite(a3_mat):
         mineig = numpy.min(numpy.real(la.eigvals(a3_mat)))
-        a3_mat += identity * (-mineig * kiter ** 2 + spacing)
+        a3_mat += identity * (-mineig * kiter**2 + spacing)
```
berland, line 633: Is this with newest black?

berland (approved):

src/fmu/tools/sensitivities/design_distributions.py
jcrivenaes: Seems so. Black is a moving target...

PR #173, opened 2023-01-12
Add rmsimport

jcrivenaes:
Inside RMS: Make it possible to load an internal module and solve some challenges (behind the curtain) that RMS has:
* Automatically add rmsproject/pythoncomp to `sys.path`
* Make it possible to edit a library module in RMS and the module import will auto-refresh
* Solve various issues with actual python extension inside RMS (e.g. `some.py_1`)

Say that "mylib.py" is a local library inside the RMS project:

```
import fmu.tools as ftools
mymodule = ftools.rms.import_localmodule(project, "mylib")
mymodule.somefunction(...)

```

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -13,6 +13,7 @@ Contents:
    qcproperties
    qcreset
    create_rft_ertobs
+   rms_import_localmodule
    rmsvolumetrics2csv
    ensembles
    contributing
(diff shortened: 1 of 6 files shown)


tnatt (approved):
Can't wait to use this 👍  nice functionality

PR #176, opened 2023-01-16
Add rename_rms_scripts

mferrera:
This PR adds a new script to fix some Python scripting annoyances in RMS, as well as gather useful information about said scripts. In particular it allows one to fix all non-`.py` file extensions and gather a list of scripts that are not currently used in any RMS workflow. It achieves this by parsing the `pythoncomp/.master` file, choosing the `instance_name` as the source of truth (the name displayed in RMS), and renaming the linked file on disk to this. It takes some precautions to avoid overwrites and includes the ability to take a backup.

A current limitation is that it ignores subdirectories in the `pythoncomp` directory.

This PR is being made at the suggestion of @jcrivenaes who has already done some testing with it. In includes fixes to feedback received from Jan already.


src/fmu/tools/rms/rename_rms_scripts.py
```diff
@@ -0,0 +1,520 @@
...
+        help="Print out a list of Python scripts that tracked by RMS but do not"
+        " appear to be used in any workflow",
+    )
+    parser.add_argument(
+        "-p",
+        "--pep8",
+        action="store_true",
+        help="Print out a list of Python scripts with PEP8 non-compliant"
+        " instance names",
+    )
```
jcrivenaes, line 46: Perhaps it should be clear when using these options, no files will be changed?

src/fmu/tools/rms/rename_rms_scripts.py
```diff
@@ -0,0 +1,520 @@
...
+    for iname in unused:
+        entry = master.get_entry(iname)
+        print(
+            f"""
+- instance_name:      {entry["instance_name"]}
+- standalonefilename: {entry["standalonefilename"]}"""
+        )
+    print("It's safest to double check and delete them manually.")
+
```
jcrivenaes, line 478: ... when inside RMS

src/fmu/tools/rms/rename_rms_scripts.py
mferrera: Added a `-t` flag for test runs with a description that indicates files will only _not_ change when it's flagged

PR #177, opened 2023-01-18
Reintroduce mypy and linting, tune GH actions

jcrivenaes:


.github/workflows/codecov.yml
```diff
@@ -6,33 +6,21 @@ on:
...
-
-    - name: Generate coverage report
-      run: |
-        git clone --depth 1 https://github.com/equinor/xtgeo-testdata ../xtgeo-testdata
-        pytest tests --disable-warnings --cov=fmu --cov-report=xml
-
-    - name: Upload coverage to Codecov
-      run: |
-        bash <(curl -s https://codecov.io/bash)
+      - uses: actions/checkout@v1
```
mferrera, line 14: Is there a reason for pinning to `checkout@v1` rather than `v3`, which is the latest big version I believe?

mferrera (approved):
LGTM

.github/workflows/codecov.yml
jcrivenaes: Not any particular; `v1` works so I haven't tested `v3`

PR #178, opened 2023-01-23
Add path key to import_localmodule()

jcrivenaes:
This makes it possible to add an external path as a local library

### DIFF
--- a/src/fmu/tools/rms/import_localmodules.py
+++ b/src/fmu/tools/rms/import_localmodules.py
@@ -44,36 +44,26 @@ def _detect_pyfile(path, module_root_name):
 
 
 @no_type_check
-def import_localmodule(project, module_root_name):
-    """Import a library module in RMS which exists inside the RMS project.
+def _detect_module_or_package(path, module_root_name):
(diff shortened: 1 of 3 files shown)


src/fmu/tools/rms/import_localmodules.py
```diff
@@ -109,3 +99,69 @@ def import_localmodule(project, module_root_name):
...
+def _localmodule_outside_rms(module_root_name, path):
+    """Import a library module in RMS which exists outside the RMS project."""
+
+    if not _detect_module_or_package(path, module_root_name):
+        raise ValueError(f"Cannot detect module {module_root_name}. Check spelling etc")
+
+    # Now empty sys.path for this module, allowing a refresh when library is modified:
+    sysm = sys.modules.copy()
+    for key, val in sysm.items():
+        if module_root_name in str(val):
```
tnatt, line 114: could this be checked against the key instead? it may be safer.

jcrivenaes: Did also extend documentation

tnatt (approved):
LGTM

PR #188, opened 2023-06-26
Minimize rename_rms_scripts interface

### LINKED ISSUE #187: rename_rms_scripts: make CLI interface better
The defaults, flags, logging, and printing are a bit all over the place and not very good.

mferrera:
Removes some cli options that unnecessarily complicated invocation

Resolves #187


jcrivenaes (approved):
LGTM

PR #195, opened 2023-11-22
fix bug for uncorrelated symmetric pert distrbutions

tralsos:
This bug was fixed for correlated pert distributions earlier, but was not changed for correlated pert distribution. Identical fix for these.


mferrera (approved):
👍

PR #199, opened 2023-12-04
Adjust pandas pin, test against Python 3.12

### LINKED ISSUE #198: Set pandas pin to lower bound
Somehow pandas ended up pinned >=2. It should be >=0.21.

mferrera:
Resolves #198 

Komodo is still locked to a pandas version less than 2 for the moment.

### DIFF
--- a/.github/workflows/fmu-tools.yml
+++ b/.github/workflows/fmu-tools.yml
@@ -19,17 +19,33 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11"]
-        pandas-version: ["0.21.0", "1.*"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        # 1.1.3 for Python 3.8 in RMS up to 14.2
+        # 1.5.3 for Komodo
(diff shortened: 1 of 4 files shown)


jcrivenaes (commented):
RMS version 14.2 with RMS 3.11 runs with pandas 2.0.2, so we need to ensure that xtgeo works with pandas > 2

.github/workflows/fmu-tools.yml
```diff
@@ -19,17 +19,30 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11"]
-        pandas-version: ["0.21.0", "1.*"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        # 1.1.3 for Python 3.8 in RMS up to 14.2
+        # 1.5.3 for Komodo
+        pandas-version: ["1.1.3", "1.5.3", "2.*"]
```
mferrera, line 25: I have pandas `2.*` in the test matrix so it will test it against all Python versions, but `2.0.2` could be specified for Python 3.11 in addition to this (2.* will take the latest version)

jcrivenaes (approved):
LGTM

PR #202, opened 2024-02-08
Update example code in docs

dafeda:
fmu.config does not exist and we get error
unless xlsx is opened in binary mode.

### DIFF
--- a/docs/fmudesign.rst
+++ b/docs/fmudesign.rst
@@ -132,10 +132,9 @@ Alternatively from a notebook:
 
 .. code-block:: python
 
-    from fmu.config import oyaml as yaml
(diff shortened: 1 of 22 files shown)


berland:
> fmu.config does not exist

?

```
[havb@st-linrgsn110:~]$ source /prog/res/komodo/stable/enable
(2024.02.05-py38) [havb@st-linrgsn110:~]$ python
Python 3.8.18 (default, Oct 24 2023, 08:42:25) 
[GCC 9.1.1 20190605 (Red Hat 9.1.1-2)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import fmu
>>> from fmu import config
>>> from fmu.config import oyaml as yaml
>>> 
```

PR #203, opened 2024-02-18
Move to ruff

### LINKED ISSUE #201: Move to ruff
black is currently failing, let's move to ruff as with other repositories.

tnatt:
Resolves #201


CONTRIBUTING.md
```diff
@@ -62,17 +62,10 @@ In addition:
...
 
 ```sh
-  python -m flake8 mycode.py
-```
-
-The pylint is rather strict and sometimes exceptions are needed.
-A targeted ``.pylintrc`` file is in the project root.
-
-```sh
-  python -m pylint mycode.py
```
mferrera, line 75: I actually think it's fine to keep `pylint`. It doesn't take a big role in the CI cycle because it's extremely pedantic but still checks a great number of things that ruff does not. A good code barometer but not a bar to meet.

However, I don't think we've been using it, so no problem to remove.

src/fmu/tools/sensitivities/_excel2dict.py
```diff
@@ -84,8 +84,9 @@ def _find_geninput_sheetname(input_filename):
 
     if len(general_input_sheet) > 1:
         raise ValueError(
-            "More than one sheet with general input"
-            "Sheetnames are {} ".format(general_input_sheet)
+            "More than one sheet with general input. Sheetnames are {} ".format(
+                general_input_sheet
+            )
```
mferrera, line 89: You might consider making these regular format strings,

`f"More than one sheet with general input. Sheetnames are {general_input_sheet} "`

I'm surprised ruff doesn't catch it. Maybe there's a reason for the method call?

src/fmu/tools/sensitivities/_excel2dict.py
```diff
@@ -120,8 +121,9 @@ def _find_onebyone_defaults_sheet(input_filename):
             default_values_sheet.append(sheet)
     if len(default_values_sheet) > 1:
         raise ValueError(
-            "More than one sheet with default values"
-            "Sheetnames are {} ".format(default_values_sheet)
+            "More than one sheet with default values. Sheetnames are {} ".format(
+                default_values_sheet
+            )
```
mferrera, line 126: Here too with f-string

mferrera (approved):
Nice job. Some comments for your consideration.

I lean toward removing pylint since presently we are not using it.

src/fmu/tools/sensitivities/_excel2dict.py
tnatt: or maybe be specific and add the `UP032` https://docs.astral.sh/ruff/rules/f-string/ ?

mferrera: I'm not sure we can completely use `UP` because sometimes we still need to fallback on Python 3.8 type annotations, even with importing future annotations. It might be reasonable to add `UP` completely and exclude this (or other) particular rules.

mferrera: Either way, maybe that should be a separate PR, so we can keep the configurations in-sync across all repositories 😄

tnatt: Ok, sounds reasonable with separate PR 🙂  then I'll merge this now!

PR #204, opened 2024-02-19
Bugfix in `rename_rms_scripts`

tnatt:
Fix bug in `rename_rms_scripts` that caused the command line utility to fail with message:
`AttributeError: 'Namespace' object has no attribute 'test'`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,6 +54,7 @@ tests = [
     "pytest-cov",
     "pytest-runner",
     "pytest-xdist",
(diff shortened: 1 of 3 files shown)


mferrera (approved):
🚀 👍 

If you're up for it please make 1.14.1 too and we'll get it into Komodo

tnatt:
> 🚀 👍
> 
> If you're up for it please make 1.14.1 too and we'll get it into Komodo

I'll make it happen 👍

PR #205, opened 2024-03-01
Format with latest ruff

mferrera:


KetilN (commented):
LGTM. 

(With one exception, it is only about whitespace formatting in the header.)

PR #209, opened 2024-04-04
Add general Sw functions as a reusable library

jcrivenaes:
The memo here is relevant to look at, and the PDF is also included in the PR. https://www.overleaf.com/read/sntvjfrtvvzh#b13262


src/fmu/tools/_common.py
```diff
@@ -50,3 +52,17 @@ def force_stop(string):
         print()
 
         sys.exit(string)
+
+
+def preserve_cwd(func):
```
tnatt, line 57: I think the `monkeypatch.chdir` would do the same as this function

jcrivenaes: Thanks. Will add as an issue

PR #211, opened 2024-04-23
Update fmutools-publish-pypi.yml

jcrivenaes:
Use trusted pypi publishing, #210

Not sure how to test this prior to merge?

### DIFF
--- a/.github/workflows/fmutools-publish-pypi.yml
+++ b/.github/workflows/fmutools-publish-pypi.yml
@@ -25,9 +25,12 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: 3.11
-          
-      - name: Install dependencies
-        run: pip install --upgrade setuptools build
+
+      - name: Update pip
+        run: pip install pip --upgrade
+        
+      - name: Install build dependencies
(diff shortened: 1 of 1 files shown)


.github/workflows/fmutools-publish-pypi.yml
```diff
@@ -1,31 +1,36 @@
 name: Publish Python 🐍 distributions 📦 to PyPI
 
 on:
+  pull_request:
+    branches: [main]
```
mferrera, line 5: With the conditional `if: github.event_name == 'release'` on the job configuration will this do anything? I think since this is a pure Python package it would be OK to skip a build on PR

mferrera (approved):
Just one comment for your consideration

.github/workflows/fmutools-publish-pypi.yml
jcrivenaes: I will edit

mferrera:
You may still need/want to have a 

```yml
- name: Install build dependencies
  run: |
    pip install -U pip
    pip install build
```
          step

Issue #216, opened 2024-08-29
The `_roxar` library is not available in RMS 14.2


jcrivenaes:
Some tests fails:

`FAILED tests/rms/test_qcreset_in_roxenv.py::test_qcreset - NameError: name '_roxar' is not defined`

The `_roxar` shall be replaced, probably with `_rmsapi` if version never than 14.0

Issue #215, opened 2024-08-29
The function `extract_grid_zone_tops()` sometimes fails after `xtgeo 3.9.2`


jcrivenaes:
```
  File "/private/jriv/work/git/fmu-tools/src/fmu/tools/extract_grid_zone_tops.py", line 118, in extract_grid_zone_tops
    dfs.append(dframe.drop(columns=["TopName", "Q_INCL", "Q_AZI"]))
```

After `xtgeo 3.9.2`, these tmp columns may not exist. The recipe is to add `error="ignore"` to the `drop()` function

PR #217, opened 2024-08-29
BUG: fix 2 small bugs wrt to XTGeo 3.9.2 and Roxar/RMS API 

jcrivenaes:
Solves #216 and #215

### DIFF
--- a/src/fmu/tools/extract_grid_zone_tops.py
+++ b/src/fmu/tools/extract_grid_zone_tops.py
@@ -104,7 +104,7 @@ def extract_grid_zone_tops(
             .map(xtg_well.get_logrecord(xtg_well.zonelogname))
             .fillna("Outside")
         )
-        dfs.append(dframe.drop(columns=["TopName", "Q_INCL", "Q_AZI"]))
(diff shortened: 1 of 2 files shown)


src/fmu/tools/extract_grid_zone_tops.py
```diff
@@ -7,18 +7,24 @@
...
-        def Project(self):
-            pass
+        class MockRoxar:
+            @property
+            def Project(self):
+                pass
 
-    _roxar = MockRoxar()
+        _roxar = MockRoxar()
+        Project = _roxar.Project
```
tnatt, line 23: It seems to me that all of this is related to type hinting the `project` argument in the function. Could we just remove it and put `Any` as type instead as we do in the `qcreset.py` 🙂 ? for simplicity?

jcrivenaes: I was thinking of that, but left it as it was your code initially; :-) I can change it to Any

tnatt (approved):
LGTM 🙂

PR #220, opened 2024-09-09
BLD, DOC: Pin sphinx-autodoc-typehints, fix incorrect usage example

### LINKED ISSUE #219: fmudesign: docs give incorrect usage for notebooks
The docs recommend

https://github.com/equinor/fmu-tools/blob/35af710fc33eb53c1f9304e09ae95600bcc71e0f/docs/fmudesign.rst?plain=1#L131-L139

This usage is not correct. It should be

```python
input_dict = excel2dict_design("../input/config/design_input_example1.xlsx")
```

where `excel2dict_design` expects a string, not a file handle.

mferrera:
Resolves #219


KetilN (approved):
LGTM

PR #233, opened 2024-10-24
Sample discrete using latin hypercube sampling

dafeda:


src/fmu/tools/sensitivities/design_distributions.py
```diff
@@ -9,6 +9,7 @@
 import numpy.linalg as la
 import pandas as pd
 import scipy.stats
+from scipy.stats import qmc
```
xjules, line 12: But bear in mind that this module was not supported in scipy 1.2, which is set as minimal in `pyproject.toml`.

dafeda: Good catch. I've bumped the version.

src/fmu/tools/sensitivities/design_distributions.py
```diff
@@ -487,21 +487,57 @@ def sample_discrete(dist_params, numreals):
...
+            raise ValueError("Probabilities must sum to 1")
+
+        fractions = probabilities
+
     elif len(dist_params) == 1:  # uniform
-        values = np.random.choice(outcomes, numreals)
+        fractions = [1.0 / len(outcomes)] * len(outcomes)
+
     else:
         status = False
```
xjules, line 525: Maybe the status variable can be removed as it is not really used.

tests/sensitivities/test_design_distributions.py
```diff
@@ -209,21 +209,69 @@ def test_draw_values_loguniform():
...
 
+def test_that_discrete_matches_with_expected_probabilities():
+    n_samples = 20
+    outcomes = ["a", "b", "c"]
+    probabilities = "0.2,0.3,0.5"
+    dist_params = [",".join(outcomes), probabilities]
+    status, values = dists.sample_discrete(dist_params, n_samples)
+    assert status
+
+    from collections import Counter
```
xjules, line 220: can this be moved to the import section?

dafeda: It can but should it?

src/fmu/tools/sensitivities/design_distributions.py
dafeda: Maybe but probably not in this PR.

tests/sensitivities/test_design_distributions.py
xjules: Yes it should. We tend to do local imports only when we have some circular imports.

dafeda:
Closing this because LHS will be introduced for all distributions in a single PR to limit the number of times we add breaking changes:

https://github.com/equinor/fmu-tools/pull/246

PR #235, opened 2024-10-25
Simplify draw_values_uniform and add tests

dafeda:


tests/sensitivities/test_design_distributions.py
```diff
@@ -183,15 +183,49 @@ def test_draw_values_lognormal():
...
+    n_samples = 1000
+
+    cov_matrix = [[1.0, correlation], [correlation, 1.0]]
+    normal_scores = rng.multivariate_normal([0, 0], cov_matrix, size=n_samples)
+
+    values1 = dists.draw_values_uniform([0, 100], n_samples, normal_scores[:, 0])
+    values2 = dists.draw_values_uniform([10, 50], n_samples, normal_scores[:, 1])
+
+    actual_correlation = np.corrcoef(values1, values2)[0, 1]
+    assert (actual_correlation - correlation) < 0.1
```
berland, line 198: Can this be flaky? Has it been tested with e.g `pytest-repeat` and `pytest --count=1<manyzeros>`?

dafeda: Good question! I did in fact test it with count=10000:

`10000 passed in 11.30s`

berland (approved):

PR #236, opened 2024-10-28
Add Metadata sheet

dafeda:
Shows which verison of fmu-tools was used and when it was created.

image

### DIFF
--- a/src/fmu/tools/sensitivities/create_design.py
+++ b/src/fmu/tools/sensitivities/create_design.py
@@ -3,11 +3,13 @@ and DESIGN_KW in FMU/ERT.
 """
 
 from collections import OrderedDict
+from datetime import datetime
 from pathlib import Path
 
 import numpy as np
 import pandas as pd
 
+import fmu.tools
 from fmu.tools.sensitivities import design_distributions as design_dist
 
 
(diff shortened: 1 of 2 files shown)


src/fmu/tools/sensitivities/create_design.py
```diff
@@ -194,6 +196,18 @@ def to_xlsx(
         defaults.to_excel(
             xlsxwriter, sheet_name=defaultsheet, index=False, header=False
         )
+
+        version_info = pd.DataFrame(
+            {
+                "Description": ["Created using fmu-tools version:", "Created on:"],
+                "Value": [
+                    fmu.tools.__version__,
+                    datetime.now().strftime("%Y-%m-%d %H:%M:%S"),
```
berland, line 205: Recommend to use `datetime.now().isoformat(sep=" ", timespec="seconds")`

tests/sensitivities/test_create_design.py
```diff
@@ -161,6 +164,24 @@ def test_generate_onebyone(tmpdir):
...
+    assert (diskmetadata.columns == ["Description", "Value"]).all()
+    assert diskmetadata["Description"].iloc[0] == "Created using fmu-tools version:"
+    assert diskmetadata["Value"].iloc[0] == fmu.tools.__version__
+    assert diskmetadata["Description"].iloc[1] == "Created on:"
+
+    # For the timestamp, we can't check the exact value since it will differ
+    # each time the test runs. Instead, we can verify it's a valid datetime string
+    # by attempting to parse it
+    timestamp = diskmetadata["Value"].iloc[1]
+    try:
```
berland, line 180: assert instead that `datetime.fromisoformat()` does not err

berland: You could mock datetime.now() if you want something exact to assert, but I am fine with this :)

src/fmu/tools/sensitivities/create_design.py
dafeda: Thanks, fixed.

tests/sensitivities/test_create_design.py
dafeda: Thanks, I'm now using `datetime.fromisoformat()` instead. Don't want to do mocking though! 😬

berland (approved):
👍🏻 👍🏻

PR #237, opened 2024-10-28
Simplify draw_values_lognormal and add tests

dafeda:


src/fmu/tools/sensitivities/design_distributions.py
```diff
@@ -197,23 +197,24 @@ def draw_values_lognormal(dist_parameters, numreals, rng, normalscoresamples=Non
...
-            distribution = scipy.stats.lognorm(s=sigma, scale=exp(mean))
-            values = distribution.rvs(size=numreals, random_state=rng)
-    else:
+
+    if not status:
         raise ValueError(msg)
+
+    mean = float(dist_parameters[0])
+    sigma = float(dist_parameters[1])
+    if normalscoresamples is not None:
```
xjules, line 207: I guess we can't rename this to `normal_score_samples`?

dafeda: Want to change as little as possible pr. PR.

xjules (approved):
Let's ship it. Although I think this would require more refactoring to be more shinny :smile: 
Like  being more explicit instead of have list of `dist_parameters`...

PR #239, opened 2024-10-30
Use RandomState instead of global random seed

dafeda:


berland (approved):
I trust you have inserted `rng` everywhere :)

dafeda:
> I trust you have inserted `rng` everywhere :)

To the best of my ability... 😬

PR #241, opened 2024-10-31
Pass rng to background and add test

dafeda:
Changed path in `extern_file` column from a non-existing file to "doe1.xlsx".

image


tralsos (approved):
Thanks, looks good.

PR #244, opened 2024-10-31
domainconversion: inherit inline/xline from input cube

### LINKED ISSUE #240: domain-conversion: Inherit inline/xline numbering from input cube
The output cube should inherit the inline/xline numbering from the input when doing domain conversion of a cube.

jcrivenaes:
Closes #240


hflesche (approved):
LGTM.

PR #246, opened 2024-11-04
Use Latin Hypercube Sampling with Iman-Conover transformation

dafeda:
# Approach

- Switches from Monte Carlo to Latin Hypercube sampling
- Uses Iman-Conover instead of directy Cholesky to induce correlations. Iman-Conover preserver marginal distributions.
- Works directly with correlation matrices instead of covariance matrices, simplifying the code
- For uncorrelated variables, uses `LatinHypercube` with dimension=1, equivalent to stratified sampling

## Comparisons

Here I compare MC with LHS with n=30 for a few distributions. I've also added results from @Risk, but note that the seed is not the same.

image
image

image
image

## Simulating correlated variables with the Cholesky factorization

https://mlisi.xyz/post/simulating-correlated-variables-with-the-cholesky-factorization/

## Allowing correlations for discrete variables

https://github.com/equinor/fmu-tools/issues/151

https://kb.palisade.com/index.php?pg=kb.page&id=72


pyproject.toml
```diff
@@ -41,7 +41,7 @@ dependencies = [
     "openpyxl>=2.6",
     "pandas>=0.21",
     "pyyaml>=5.3",
-    "scipy>=1.2",
+    "scipy>=1.7",
```
mferrera, line 44: Just a heads-up: we have to support scipy 1.5.3 at minimum in this package due to it being included in RMS environments. All RMS versions below 14.2.2 (which is the newest and still not widely adopted by assets) lock scipy to 1.5.3, a vendor limitation we can't relax

dafeda: Thanks!
Do you know if `fmudesign` is used from within RMS?

mferrera: To my knowledge it's not used directly but @jcrivenaes may need to correct me

dafeda: If not, could we consider moving to a separate repo?

jcrivenaes: > To my knowledge it's not used directly but @jcrivenaes may need to correct me

Not to my knowledge either

mferrera: > If not, could we consider moving to a separate repo?

I think this is completely fine if Jan and Trine think so as well, so long as we hang a link from the design docs here to its new location

jcrivenaes: OK with me

tralsos: Yes, I think that should be ok. Quite sure it's not used within RMS.

dafeda:
Moved to `semeio`: https://github.com/equinor/semeio/pull/660

PR #249, opened 2024-11-06
Do not sample from multivariate_normal in test

dafeda:
Instead, add correlated PARAM5 ~ TruncatedNormal(3, 1, 1, 5) and PARAM6 ~ Uniform(0, 1) to design_input_background.xlsx. We don't want tests that call rng.multivariate_normal directly as that will change when we do latin hypercube sampling instead. Add check that lower bound must be less than upper bound in truncated normal


tralsos (approved):
LGTM

PR #250, opened 2024-11-08
Deprecate calc_tornadoinput

dafeda:
Use TornadoPlotterFMU from webviz instead


tralsos (commented):
So this will add a deprecation warning, but not remove it yet ?

dafeda:
> So this will add a deprecation warning, but not remove it yet ?

That is correct.
If we for example run `test_calc_tornado.py`, we get the following warning:

` /Users/FCUR/git/fmu-tools/tests/sensitivities/test_calc_tornado.py:59: DeprecatedWarning: calc_tornadoinput is deprecated. Use TornadoPlotterFMU from webviz instead`

tralsos (approved):

PR #252, opened 2024-11-08
Test summarize design

dafeda:


src/fmu/tools/sensitivities/_designsummary.py
```diff
@@ -34,7 +40,20 @@ def summarize_design(filename, sheetname="DesignSheet01"):
 
     """
 
-    # Initialisation of dataframe to store results
+    # Read design matrix
+    if str(filename).endswith(".xlsx"):
+        dgn = pd.read_excel(filename, sheetname, engine="openpyxl")
```
xjules, line 45: I'm sure that you don't want to change the naming conventions but `dgn` is not a great name to be honest.

src/fmu/tools/sensitivities/_designsummary.py
```diff
@@ -34,7 +40,20 @@ def summarize_design(filename, sheetname="DesignSheet01"):
...
+        dgn.dropna(axis=0, how="all", inplace=True)
```
xjules, line 46: Hm, this drops rows (realization) if a parameter entry is missing. Shouldn't we drop the column (the parameter) instead? Not sure just asking ....

dafeda: I notice that I've deleted the original comment that explains this line.
I think it makes sense to drop the row.

tralsos (approved):

PR #253, opened 2024-11-12
Sample simgrid seismic as points

### LINKED ISSUE #223: Make a function that for sampling of grid regions and seismic attributes to point like format for ERT usage
Today, some "attribute maps" are made in RMS in the sim2seis workflow. In accordance with sim2seis being independent of vendor tools like RMS, we wants this functionality to be taken out.

This is triggered by https://github.com/equinor/fmu-sim2seis/issues/27

jcrivenaes:
This will potentially remove several jobs/script in RMS and replace them with one python script. It also makes it possible to run this independent of RMS, hence it is linked to  https://github.com/equinor/fmu-sim2seis/issues/27

Closes #223 

https://link.excalidraw.com/readonly/LzQYxTHfLcviR8F7GLRW

### DIFF
--- a/src/fmu/tools/utilities/sample_attributes.py
+++ b/src/fmu/tools/utilities/sample_attributes.py
@@ -1,0 +1,184 @@
+"""Sample attributes on grid resolution as dataframe points sets.
+
+This usage is for setting attributes on grid resolution, e.g. a seismic attribute
+(from a map) combined with a region parameter from the same 3D grid.
+
+This is targeted to the "sim2seis" workflow in FMU.
+"""
+
+from __future__ import annotations
+
+import logging
+import pathlib
+import tempfile
+from enum import Enum
+from typing import TYPE_CHECKING, Any
+
+import numpy as np
+import xtgeo
+
+if TYPE_CHECKING:
+    import pandas as pd
+
+# custom logger for this module
+logger = logging.getLogger(__name__)
+
+
+class Attrs(str, Enum):
+    OBS = "OBS"
+    OBS_ERROR = "OBS_ERROR"
+    REGION = "REGION"
+    VALUES = "VALUES"
+
+
+class Position(str, Enum):
+    TOP = "top"
+    CENTER = "center"
+    BASE = "base"
+
+
+def _get_layer(
+    position: tuple[str, Position], grid: xtgeo.Grid, zone: xtgeo.GridProperty
+) -> int:
+    """Get the layer index (base=1!) based on the position and zone."""
+    zone_name, where = position
+
+    position_values = [pos.value for pos in Position]
+    if where not in position_values:
+        raise ValueError(f"Position shall be one of: {position_values}")
+
+    top_value = 1
+    center_value = round(0.5 * (grid.nlay + 1))
+    base_value = grid.nlay
+
+    if zone and zone_name:
+        if zone_name not in zone.codes.values():
+            raise ValueError(f"Zone name '{zone_name}' not found in the grid.")
+
+        nzone = {v: k for k, v in zone.codes.items()}.get(zone_name)
+        logger.debug("Zone index to apply for %s is: %s", zone_name, nzone)
+
+        indices = np.where(zone.values == nzone)
+        upper, lower = indices[2].min() + 1, indices[2].max() + 1  # base=1
+
+        top_value = upper
+        center_value = round(0.5 * (upper + lower))
+        base_value = lower
+
+    if Position.TOP in where:
+        return top_value
+
+    if Position.BASE in where:
+        return base_value
+
+    return center_value  # default is center
+
+
+def _dataframe_from_surface(
+    surface: xtgeo.RegularSurface, newname: str | None = None, debug: bool = False
+) -> pd.DataFrame:
+    """Get the data frame via xtgeo Points (makes it easier to debug)"""
+
+    points = xtgeo.points_from_surface(surface)
+    points.zname = newname if newname else Attrs.VALUES.value
+    df = points.get_dataframe()
+    if debug:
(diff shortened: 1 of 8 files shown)


src/fmu/tools/utilities/sample_attributes.py
```diff
@@ -0,0 +1,162 @@
...
+        if zone_name not in zone.codes.values():
+            raise ValueError(f"Zone name '{zone_name}' not found in the grid.")
+
+        nzone = {v: k for k, v in zone.codes.items()}.get(zone_name)
+        logger.debug("Zone index to apply for %s is: %s", zone_name, nzone)
+
+        indices = np.where(zone.values == nzone)
+        upper, lower = indices[2].min() + 1, indices[2].max() + 1  # base=1
+
+        return {"top": upper, "base": lower}.get(where, round(0.5 * (upper + lower)))
```
tnatt, line 48: We should check if the input `where` is within the three options `top/base/center` and raise error if not. And it would be better to be more explicit in this return to understand that the option `center` returns `round(0.5 * (upper + lower)`

src/fmu/tools/utilities/sample_attributes.py
```diff
@@ -0,0 +1,162 @@
...
+        attribute_error_sampled, newname=Attrs.OBS_ERROR.value, debug=debug
+    )
+    dataframe[Attrs.OBS_ERROR.value] = df_err[Attrs.OBS_ERROR.value]
+
+    df_region = None
+    if region:
+        region_sampled = xtgeo.surface_from_grid3d(
+            grid, template=template, where=layer, property=region
+        )
+        region_sampled.to_file("/tmp/region.gri")
```
tnatt, line 155: Was this for testing purpose?

src/fmu/tools/utilities/sample_attributes.py
```diff
@@ -0,0 +1,162 @@
...
+        df_region = _dataframe_from_surface(
+            region_sampled, newname=Attrs.REGION.value, debug=debug
+        )
+        dataframe[Attrs.REGION.value] = df_region[Attrs.REGION.value]
+
+    # remove lines without NaN values, if any
```
tnatt, line 161: don't need this comment I think 🙂 

```suggestion
    # remove lines with NaN values, if any
```

tests/utilities/test_attributes_sim2seis.py
```diff
@@ -0,0 +1,144 @@
...
+        attr_surface,
+        attribute_error=attr_surface_error,
+        region=region,
+        position=("Valysar", "top"),
+        zone=zone,
+        debug=False,  # this is a 'developer' kwargs to the function
+    )
+    assert Attrs.OBS.value in df.columns
+    assert Attrs.OBS_ERROR.value in df.columns
+    assert np.mean(df.OBS.to_numpy()) == pytest.approx(
```
tnatt, line 72: Since region is input we should assert that `REGION` is in the columns as well

tnatt, line 72: Maybe the following is easier to read
`assert np.isclose(df["OBS"].mean(), attr_surface.values.mean(), rtol=0.05)`

tests/utilities/test_attributes_sim2seis.py
```diff
@@ -0,0 +1,144 @@
...
+        attr_surface.values.mean(), rel=0.05
+    )
+    assert np.max(df.REGION.to_numpy()) == 7.0
```
tnatt, line 75: Consider using built in pandas method here for readability 

`df["REGION"].max()`

src/fmu/tools/utilities/sample_attributes.py
```diff
@@ -0,0 +1,162 @@
...
+
+    template = xtgeo.surface_from_grid3d(
+        grid,
+        template="native",  # used to appromimate the native grid resolution
+        where=layer,
+        property="i",
+    )
+    # prepare the attribute and error to the grid
+    attribute_sampled = template.copy()
+    attribute_error_sampled = template.copy()
```
tnatt, line 123: Do we need these copies here or could we just use the template object to save memory?

src/fmu/tools/utilities/sample_attributes.py
```diff
@@ -0,0 +1,162 @@
...
+        **kwargs: Additional keywords (developer settings).
+
+    Returns:
+        pd.Dataframe: Points with the sampled attributes and attributes combined.
+    """
+    logger.info("Sampling attributes on grid resolution as points set.")
+
+    debug = kwargs.get("debug", False)
+
+    layer = _get_layer(position, grid, zone)
```
tnatt, line 112: We should add a check of the `position` type before entering here. That it actually is a tuple with two values, and that the last item is a valid 'where' 🙂

tnatt (approved):
I am not too familiar with the SIM2SEIS workflow, but the functionality looks fine to me 🙂 
Added some small comments for your consideration.

src/fmu/tools/utilities/sample_attributes.py
jcrivenaes: upsi, yes

tests/utilities/test_attributes_sim2seis.py
jcrivenaes: Perhaps this is best:
`assert df.OBS.mean() == pytest.approx(attr_surface.values.mean(), rel=0.05)`

I think `np.isclose` is better for numpy vectors (here we just have scalars)

src/fmu/tools/utilities/sample_attributes.py
jcrivenaes: Think we need it (since xtgeo do "inplace" operations, not returning the object)

PR #258, opened 2024-11-19
Remove old seed deprecation warning

dafeda:


tralsos (commented):
Yes, it's a warning that has been there for a while. Perhaps the option to use "seeds" instead of "rms_seeds" should also be removed ? (line 240) This namechange came when we also introduced to possibility to set seed value for fmudesign, and it was chosen to use "rms_seeds" and "dist_seed" for the two different types of seeds.

dafeda:
> Yes, it's a warning that has been there for a while. Perhaps the option to use "seeds" instead of "rms_seeds" should also be removed ? (line 240) This namechange came when we also introduced to possibility to set seed value for fmudesign, and it was chosen to use "rms_seeds" and "dist_seed" for the two different types of seeds.

Make sense to remove the if statement dealing with "seeds" I think, thanks.

tralsos (commented):
Not sure how this is normally done. But I expect some project still have a general_input sheet where they have "seeds" instead of "rms_seeds". Would it be better if "seeds" is found in general_input sheet, to let it fail with the message that they should change this to "rms_seeds"?

tralsos (approved):

PR #264, opened 2024-12-05
Generate new petro logs from original logs as part of an AHM workflow in FMU

oddvarlia:
Purpose: 
RMS python script to create new petro logs from original petro logs with one new log per facies. 
The new logs will get name in the form:  "faciesname_petroname" where the new log has the same values as the original log with name "petroname" except that all values belonging to other facies than "faciesname" is det to undefined.

The new logs created will be used to condition petrophysical realisations where only one facies is present for all grid cells in the zone instead of using a facies realization as input when simulating petrophysical realization. This is relevant for workflows where both facies and petrophysical parameters are used as field parameters in assisted history matching with ERT.

Closes: #265


src/fmu/tools/rms/generate_bw_per_facies.py
```diff
@@ -0,0 +1,102 @@
...
+def create_bw_per_facies(
+    project,
+    grid_name: str,
+    bw_name: str,
+    original_petro_log_names: List[str],
+    facies_log_name: str,
+    facies_code_names: Dict[int, str],
+    debug_print: bool = False,
+) -> None:
+    """Script to be used as RMS python job to create new petrophysical
```
jcrivenaes, line 17: Perhaps, instead of "Script to be used...":
"Function to be imported and applied in a RMS python job..."

src/fmu/tools/rms/generate_bw_per_facies.py
```diff
@@ -0,0 +1,102 @@
...
+        Create the blocked well logs to be used to condition
+        petrophysical realizations where all grid cells are
+        assumed to belong to only one facies. This script will
+        not modify any of the original logs, only create new logs where only
+        petrophysical log values for one facies is selected
+        and all other are set ot undefined.
+
+    Input
+
+        grid model name,
```
jcrivenaes, line 33: It would be better to use standard "google" docstring style here, see example here:
https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

src/fmu/tools/rms/generate_bw_per_facies.py
```diff
@@ -0,0 +1,102 @@
...
+        well.set_dataframe(df)
+        if debug_print:
+            print(f"Well:  {well.name}")
+            print(f"All logs:  {well.lognames_all}")
+            print("Dataframe for facies log and new logs:")
+            df_updated = well.get_dataframe()
+            selected_log_names = []
+            selected_log_names.append(facies_log_name)
+            selected_log_names.extend(new_log_names)
+            print(f"{df_updated[selected_log_names]}")
```
jcrivenaes, line 85: I don't understand why you do operations under a `if debug_print` section

docs/generate_bw_per_facies.rst
```diff
@@ -0,0 +1,79 @@
...
+well logs except that all values belonging to other facies than the one for the
+created log is set to undefined. The new logs will have name of
+the form "faciesname_petroname".
+
+Example of use as RMS python job
+---------------------------------
+
+.. code-block:: python
+
+     from fmu.tools.rms.generate_bw_per_facies import create_bw_per_facies
```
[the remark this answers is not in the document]
oddvarlia: Fixed

src/fmu/tools/rms/generate_bw_per_facies.py
oddvarlia: I tried.

PR #267, opened 2024-12-12
DEP: Drop Python 3.8 support

mferrera:


jcrivenaes (approved):
👍

jcrivenaes (approved):

mferrera:
After approve change: just applied a quick additional fix to the conditions under which docs build since pandas 2.0.2 won't be present there anymore

PR #268, opened 2025-01-02
Removed a warning related to import of rmsapi

oddvarlia:
Removed a warning related to import of rmsapi  (roxapi)

### DIFF
--- a/src/fmu/tools/rms/generate_petro_jobs_for_field_update.py
+++ b/src/fmu/tools/rms/generate_petro_jobs_for_field_update.py
@@ -13,19 +13,21 @@ Summary:
 import copy
 import json
 import pprint
-import warnings
 
 import yaml
 
-try:
-    import _roxar  # type: ignore
-except ModuleNotFoundError:
+from fmu.tools import ROXAR
+
+if ROXAR:
     try:
-        import _rmsapi as _roxar  # type: ignore
-        import roxar.jobs  # type: ignore
-        import roxar.jobs.Job  # type: ignore
+        import rmsapi
+        import rmsapi.jobs as jobs
     except ModuleNotFoundError:
-        warnings.warn("This script only supports interactive RMS usage", UserWarning)
+        import roxar as rmsapi
+        import roxar.jobs as jobs
+else:
+    pass
+
 from typing import Dict, List, Tuple, no_type_check
 
 # Fixed global variables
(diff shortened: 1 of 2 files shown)


src/fmu/tools/rms/generate_petro_jobs_for_field_update.py
```diff
@@ -13,7 +13,6 @@
 import copy
 import json
 import pprint
-import warnings
 
 import yaml
```
tnatt, line 18: Will the imports here work? I think you need to import `roxar` and `rmsapi` instead of `_roxar` and `_rmsapi` ?
```python
try:
    import roxar  # type: ignore
    import roxar.jobs  # type: ignore

except ModuleNotFoundError:
    try:
        import rmsapi as roxar  # type: ignore
        import roxar.jobs  # type: ignore
```

You could potentially simplify by using the global `ROXAR` variable defined in `fmu.tools.__init__` . something like:
```python
if ROXAR:
    import rmsapi
    import rmsapi.jobs  # type: ignore
else:
    pass
```
and then use rmsapi in the script instead of roxar 🙂

tnatt (changes requested):
I tested the imports in RMS (both 14.2 and 13.2)  and it seems there still is an issue here. I added a suggestion for you consideration 🙂

src/fmu/tools/rms/generate_petro_jobs_for_field_update.py
oddvarlia: OK If we assume RMS version is new enough to use rmsapi, I will go for the last suggestion.

tnatt: The code in `fmu.tools.__init__` handles older RMS versions by importing `roxar` as `rmsapi,` so should work fine. But you could test inside RMS to make sure.

https://github.com/equinor/fmu-tools/blob/ab5c4c92f2398c7a0128662c655f06ce653408b2/src/fmu/tools/__init__.py#L10-L21

src/fmu/tools/rms/generate_petro_jobs_for_field_update.py
```diff
@@ -13,19 +13,16 @@
...
-    try:
-        import _rmsapi as _roxar  # type: ignore
-        import roxar.jobs  # type: ignore
-        import roxar.jobs.Job  # type: ignore
-    except ModuleNotFoundError:
-        warnings.warn("This script only supports interactive RMS usage", UserWarning)
+from fmu.tools import ROXAR
+
+if ROXAR:
+    import rmsapi
```
tnatt, line 22: You would need to import the `rmsapi.jobs` explicit here to avoid an AttributeError.

image

src/fmu/tools/rms/generate_petro_jobs_for_field_update.py
oddvarlia: Code is updated and should work for both RMS13.1.2 and RMS14.2. I also fixed a line in the test for generate_petro_jobs_for_field_update.py

tnatt (approved):
LGTM 👍

PR #272, opened 2025-01-23
CI: Add script for running tests from github worklow

slangeveld:
Resolves part of https://github.com/equinor/rms-sys/issues/267

Added a bash script that can be run from the github workflows to execute the `fmu-tools` unit tests towards an RMS Python Environment. The script copies the required test files, installs the test dependencies and runs the pytests. Note that the xtgeo-testdata files needed to run the tests are cached on the github actions runner som pointing to the path `XTGEO_TESTDATA_PATH` with the cached data should be sufficient to run the tests.

Tests have been updated to be able to use the testdata path `XTGEO_TESTDATA_PATH` set in the test config:
* Since the tests are to be run from a github workflow, the tests need to support xtgeo-testdata being placed at the path given by `XTGEO_TESTDATA_PATH` in the command `pytest ./tests -n 4 -vv --testdatapath $XTGEO_TESTDATA_PATH`. 

A testrun from can be seen here https://github.com/equinor/rms-sys/actions/runs/13006244695/job/36273691991

### DIFF
--- a/ci/testrmsenv.sh
+++ b/ci/testrmsenv.sh
@@ -12,15 +12,18 @@ run_tests () {
 }
 
 install_test_dependencies () {
-    pip install ".[dev]"
+    echo "Installing test dependencies..."
+    pip install ".[tests]"
 }
 
 copy_test_files () {
+    echo "Copy test files to test folder $CI_TEST_ROOT..."
     cp -r $CI_SOURCE_ROOT/tests $CI_TEST_ROOT/tests
     # Pytest configuration is in pyproject.toml
     cp $CI_SOURCE_ROOT/pyproject.toml $CI_TEST_ROOT
 }
 
 start_tests () {
+    echo "Running fmu-tools tests with pytest..."
     pytest
 }
--- a/ci/testrmsenv.sh
+++ b/ci/testrmsenv.sh
@@ -1,0 +1,26 @@
+# This shell script is to be sourced and run from a github workflow
+# when fmu-tools is to be tested towards a new RMS enviroment
+
+run_tests () {
+    copy_test_files
+
+    install_test_dependencies
+
+    pushd $CI_TEST_ROOT
+    start_tests
+    popd
+}
+
+install_test_dependencies () {
+    pip install ".[dev]"
+}
+
+copy_test_files () {
+    cp -r $CI_SOURCE_ROOT/tests $CI_TEST_ROOT/tests
+    # Pytest configuration is in pyproject.toml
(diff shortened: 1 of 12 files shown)


ci/testrmsenv.sh
```diff
@@ -0,0 +1,47 @@
...
+
+    echo "Copy fmu-tools test files to test folder $CI_TEST_ROOT..."
+    mkdir -p $CI_TEST_ROOT  
+    cp -r $PROJECT_ROOT/tests $CI_TEST_ROOT/tests
+    cp $PROJECT_ROOT/pyproject.toml $CI_TEST_ROOT
+}
+
+install_test_dependencies () {
+    echo "Installing test dependencies..."
+    pip install --upgrade --force-reinstall ".[tests]"
```
slangeveld, line 36: All tests for all of the projects are running in the same _rmsenv, so using the `--upgrade --force-reinstall` arguments to make sure to overwrite what is already installed

ci/testrmsenv.sh
```diff
@@ -0,0 +1,47 @@
...
+
+    echo "Dependencies installed successfully. Listing installed dependencies..."
+    pip list
+}
+
+run_pytest () {
+    echo "Running fmu-tools tests with pytest..."
+    pushd $CI_TEST_ROOT
+    pytest -n 4 -vv -m "not skipunlessroxar"
```
mferrera, line 45: My brain can't boolean today, but I think we _want_ to include the the roxar tests (does this do that)?

ci/testrmsenv.sh
mferrera, line 36: Future selves shall thank us for an isolated install 👍

mferrera (approved):
Looks good 👍 

For posterity there was some offline discussion about this

ci/testrmsenv.sh
slangeveld: The intention here was to exclude it, to avoid calling the RMS API to often as discussed here https://github.com/equinor/rms-sys/pull/254

But open to include it if we want that.

ci/testrmsenv.sh
slangeveld: Have updated to run on rmsvenv for each project, so that we can remove `--upgrade --force-reinstall`

ci/testrmsenv.sh
slangeveld: Saw your comment here https://github.com/equinor/xtgeo/pull/1297

Will include the roxar tests 👍

PR #278, opened 2025-03-10
FIX: Reinstate fmu.tools.volumetrics

mferrera:
This import was erroneously removed with fmudesign.


berland (approved):
Looks good. There should probably be a test that ensures this is in place.

slangeveld (approved):

PR #280, opened 2025-04-22
Updated copy_rms_param_to_ertbox.py

oddvarlia:
Need more flexible handling of zone numbers. Zone numbers defined in Zone parameter (3D RMS parameter for a grid) is defined by a dict of type` {zone_number: zone_name}` where zone_number does not have to be related to the index for the zone in the grid. Usually a grid with multiple zones has zone names from top to base and specified by a list `zone_names[index]` where index starts at 0 with N zones. Usually (when Zone parameter (3D grid parameter) is created for the grid, the zone numbers in the dict `{zone_number: zone_name}` start with zone_number = 1 and end with zone_number=N for the zone at the bottom of the grid.
This is however, not always the case. If the original grid is split into single zone grids or smaller multi-zone grids, there are a subset of the original zones in the new subgrids. The zone parameter can , however,  sometimes use the original zone_numbers.

Example:  

Original grid with 3 zones:    ZoneA, ZoneB, ZoneC  with zone list  `zone_names = ['ZoneA', 'ZoneB', 'ZoneC']`
Original zone parameter:   ` zone_code_names = {1:'ZoneA', 2: 'ZoneB',  3: 'ZoneC'}`
In this case` zone_number = zone_index + 1`

The original grid is split into smaller subgrids (single zone grids):
SubgridA contains zone 'ZoneA' with zone parameter with zone dict` zone_code_names = {1: 'ZoneA'}`
SubgridB contains zone 'ZoneB'  with zone parameter with zone dict` zone_code_names = {2: 'ZoneB'}`
SubgridC contains zone 'ZoneC'  with zone parameter with zone dict `zone_code_names = {3: 'ZoneC'}`

zone_names_list for SubgridA is` zone_names = ["ZoneA"]`   zone_index = 0 in this case and zone_number is 1
zone_names_list for SubgridB is` zone_names = ["ZoneB"]`   zone_index = 0 in this case  and zone_number is 2
zone_names_list for SubgridC is `zone_names = ["ZoneC"]`  zone_index = 0 in this case  and zone_number is 3

Conclusion: We need a mapping to define correspondence between zones in the grid and zones in the zone parameter.

Closes: #281


src/fmu/tools/rms/zone_mapping.py
```diff
@@ -0,0 +1,242 @@
...
+    grid_zone_index_dict: dict[int, int] = {}
+
+    def __init__(
+        self,
+        grid_model: Any,
+        grid: Any,
+        real_number: int = 0,
+        fmu_mode: bool = False,
+        debug_level: int = DEBUG_OFF,
+    ):
```
jcrivenaes, line 39: It appears that you first make variables as class atrributes, and you overwrite with instance attributes (.e.g `grid_model` is first a class attribute, and then becomes an instance attribite after `__init__`). I propose you remove the class attributes unless you have a spesific use for them?

src/fmu/tools/rms/zone_mapping.py
```diff
@@ -0,0 +1,242 @@
...
+            )
+            self.zone_code_names = zone_param.code_names
+        else:
+            self.zone_code_names = {1: "Zone1"}
+
+        indexer = self.grid.simbox_indexer
+        zonation = indexer.zonation
+        if debug_level >= DEBUG_VERY_VERBOSE:
+            print(f"--- Zone parameter with zone_code_names: {self.zone_code_names}")
+            print(f"--- Grid with zone names: {self.grid.zone_names}")
```
jcrivenaes, line 61: Think I have mentioned it before, but using python logging for debug messages is lightlyears better. It gives much more control and can be steered from outside while running. I can give you a brief how-to if needed.

src/fmu/tools/rms/zone_mapping.py
```diff
@@ -0,0 +1,242 @@
...
+        for zone_index in range(nzones):
+            zone_numbers.append(self.grid_zone_dict[zone_index]["zone_number"])
+        return zone_numbers
+
+    def get_zone_number(self, zone_name: str):
+        # Unique zone name for each zone code
+        for code, name in self.zone_code_names.items():
+            if zone_name == name:
+                return code
+        raise ValueError(f"Unknown zone name: {zone_name}")
```
jcrivenaes, line 138: In general (applies to this and almost all functions): If using type annotations, also indicate return type
, e.g.

`def get_zone_number(self, zone_name: str) -> int:`

src/fmu/tools/rms/zone_mapping.py
```diff
@@ -0,0 +1,242 @@
...
+        for zone_index in range(nzones):
+            number_of_layers_per_zone.append(
+                self.number_of_layers_for_zone_index(zone_index)
+            )
+            start, end = self.get_start_end_layer_for_zone_index(zone_index)
+            start_layers_per_zone.append(start)
+            end_layers_per_zone.append(end)
+        return number_of_layers_per_zone, start_layers_per_zone, end_layers_per_zone
+
+    def get_number_of_layers_per_zone(self):
```
jcrivenaes, line 202: I don't think this method is applied anywhere. It is also overlapping with `get_zone_layer_numbering`?

src/fmu/tools/rms/zone_mapping.py
```diff
@@ -0,0 +1,242 @@
...
+        return self.zone_code_names[zone_number]
+
+    def get_zone_number_for_zone_index(self, zone_index: int):
+        return self.grid_zone_dict[zone_index]["zone_number"]
+
+    def get_zone_index_for_zone_number(self, zone_number: int):
+        self.validate_zone_number_for_grid(zone_number)
+        return self.grid_zone_index_dict[zone_number]
+
+    def is_zone_number_defined(self, zone_number: int):
```
jcrivenaes, line 164: Is this used?

src/fmu/tools/rms/copy_rms_param_to_ertbox_grid.py
```diff
@@ -1296,29 +1287,29 @@ def check_geogrid_parameters(
...
+#            zone_number - 1
+#        ]  # Zonation is 0-indexed, while zone numbers are 1-indexed
+#    start_layer = nz
+#    end_layer = 0
+#    for layer_range in layer_ranges:
+#        if start_layer > layer_range.start:
+#           start_layer = layer_range.start
+#        if end_layer < layer_range.stop:
+#            end_layer = layer_range.stop
+#    return end_layer, start_layer
```
jcrivenaes, line 1312: Instead of commenting out, just remove (git will keep track)

jcrivenaes (changes requested):
Many changes and it probably works well. I propose that you remove functions that are not used, and also there are several functions that appears to do the same(?) or close to the same. Possible to refactor? Also consider usage of class_variables vs instance variables.

src/fmu/tools/rms/copy_rms_param_to_ertbox_grid.py
oddvarlia: yes

src/fmu/tools/rms/zone_mapping.py
oddvarlia: Is used in APS code, but get_zone_layer_numbering can be removed

src/fmu/tools/rms/zone_mapping.py
jcrivenaes: We could consider adressing it in own issue/PR

jcrivenaes (approved):
👍

PR #297, opened 2025-08-04
BUG: fix issue in swfunction when x value gets zero

### LINKED ISSUE #296: [swfunction] The result Sw may get masked if `X` is zero in some cells
The `swfunction` has a `x` input which is the typically the J function without the height term. 

In some places in the code this variable is the denominator, which implicitly seem to trigger that the mask of the masked Numpy array is turned True.

Tasks:
- Avoid division on zero
- Ensure that the grid's ACTNUM and mask of the returned variable is consistent.

jcrivenaes:
Closes #296


alifbe (approved):
Looks good to me 👍

PR #300, opened 2025-09-18
`swfunction` Add new method for height above FFL and refactoring

alifbe:
- Add new method for computing height above FFL (and removed invalid one)
- Refactor debug mode and instead return useful properties together with Sw. As library, `swfunction` should let users  handle the properties themselves. Note: This is breaking change.

### DIFF
--- a/tests/properties/test_swfunction.py
+++ b/tests/properties/test_swfunction.py
@@ -17,6 +17,7 @@ from fmu.tools.properties import SwFunction
         (1, -2, 13, False, "cell_center_above_ffl", 0.412536),
         (0.3, -5, 12.5, True, "cell_center_above_ffl", 0.547251),
         (0.3, -5, 12.5, False, "cell_center_above_ffl", 0.549613),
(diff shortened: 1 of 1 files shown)


src/fmu/tools/properties/swfunction.py
```diff
@@ -331,4 +309,11 @@ def compute(self, compute_method: str = "integrated") -> xtgeo.GridProperty:
...
 
-        return self._sw
+        output_props = [self._sw, self.htop, self.hbot, self.hcenter]
+        output_props_name = ["SW", "HTOP", "HBOT", "HCENTER"]
+
+        for prop, prop_name in zip(output_props, output_props_name):
+            assert isinstance(prop, xtgeo.GridProperty)
+            prop.name = self.tag + prop_name
+
+        return tuple(output_props)
```
jcrivenaes, line 319: Instead of returning a tuple, could a dict return be more user friendly and also more flexible in the future?
```
sw = sw_obj.compute("integrated")["SW"]
```

PR #372, opened 2026-04-10
ENH: add nested hybrid grid functionality

### LINKED ISSUE #355: Add functionality for nested hybrid grids
This task is based on the following workflow

- Make a region in the 3D grid that shall be refined
- Do refinement if the grid wihin this region. This involves a 'trick' which is to extend the grid in IJK coordinates, but let those coordinates overlap in real coorinates with the current grid
- Do a proper manipulation of ACTNUM for the hybrid case
- Set up NNCS to connect the hybrid grid with the original grid

jcrivenaes:
Resolves #355 

Functions for creating nested hybrid grids.

One MAINT commit also to fix some mypy issues that has emerged

## Checklist

### DIFF
--- a/docs/nestedhybridgrid.rst
+++ b/docs/nestedhybridgrid.rst
@@ -1,0 +1,149 @@
+Nested Hybrid Grids
+====================
+
+.. contents:: Table of Contents
+   :local:
+   :depth: 2
+
+Overview
+--------
+
+The ``nestedhybridgrid`` module creates **nested hybrid grids** where a
+selected region of a coarse grid is replaced by a refined (subdivided)
+sub-grid.  The two grids are merged into a single grid and connected
+through Non-Neighbour Connections (NNCs).
+
+The typical workflow is:
+
+1. Define a coarse grid and a region property that marks which cells to
+   refine.
+2. Call :func:`~fmu.tools.nestedhybridgrid.create_nested_hybrid_grid` to
+   produce the merged grid and an NNC table. You may need to export the NNC file to csv
+   at this stage.
+3. Do rescaling from the original gridmodel (e.g. a finer geogrid) to the merged grid,
+   e.g. by using software like RMS.
+4. Make another script, that computes transmissibilities with
+   :meth:`xtgeo.Grid.get_transmissibilities` passing the NNC table.
+5. Export the NNC transmissibilities for the flow simulator.
+
+Quick-start example
+-------------------
+
+The example here runs within RMS, but similar workflows can be created for file i/o.
+
+.. code-block:: python
+
+    import xtgeo
+    from fmu.tools.nestedhybridgrid import (
+        create_nested_hybrid_grid,
+    )
+
+    # Load grid and region property
+    grid = xtgeo.grid_from_roxar(project, "Simgrid")
+    region = xtgeo.gridproperty_from_roxar(project, "Simgrid", "REGION")
+
+    # Create nested hybrid grid (e.g. refine region 2 by 2×2×1)
+    merged, nnc_table = create_nested_hybrid_grid(
+        grid, region, target_region_id=2, refinement=(2, 2, 1)
+    )
+
+    # store merged grid in RMS (or file)
+    merged.to_roxar(project, "NestedHybrid")
+
+    merged.get_prop_by_name("NEST_ID")
+    nest_id.to_roxar(project, "NestedHybrid", nest_id.name)
+
+    # write the NNC pandas to disk; this will be applied for computing NNC's in the next script
+    nnc_table.write_csv("path_to_some_csv_file.csv", index=False)
+
+
+The next step is to do a rescaling from the original geogrid to the merged grid
+using e.g. the RMS tool.
+
+Further, we need to create NNC transmissibilities and generate file for flow simulator:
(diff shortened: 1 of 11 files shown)


docs/nestedhybridgrid.rst
```diff
@@ -0,0 +1,173 @@
...
+.. code-block:: text
+
+    INCLUDE
+      'NNC_HYBRID.INC' /
+
+API reference
+-------------
+
+See :doc:`fmu.tools.nestedhybridgrid` for the full auto-generated API
+reference.
```
alifbe: There is no `fmu.tools.nestedhybridgrid` document to link to. This will cause a warning when building the document.

src/fmu/tools/__init__.py
```diff
@@ -27,13 +27,17 @@
...
+from fmu.tools.nestedhybridgrid.nestedhybrid import nnc_to_gridproperty  # noqa
 
 __all__ = [
     "extract_grid_zone_tops",
     "wellzonation_vs_grid",
     "QCProperties",
     "DomainConversion",
     "sample_attributes_for_sim2seis",
+    "create_nested_hybrid_grid",
+    "nnc_to_gridproperty",
```
alifbe, line 42: We also need to expose `nnc_to_flowsimulator_input` function here.

docs/nestedhybridgrid.rst
alifbe, line 1: Some comments related to the documentation:
- For some reason, the documentation is now missing the API page.
Image
- The documentation format for `nested-hybrid-grid` is different compare to other module. In other modules, the API is not part of main documentation. 
- There is no `fmu.tools` document, so we will get warning when building documentation.

docs/nestedhybridgrid.rst
```diff
@@ -0,0 +1,176 @@
...
+
+.. code-block:: text
+
+    INCLUDE
+      'NNC_HYBRID.INC' /
+
+API reference
+-------------
+
+See :doc:`fmu.tools` for the auto-generated API pages, including
```
alifbe: There is no document `fmu.tools.rst` so this will trigger a warning during building.

src/fmu/tools/nestedhybridgrid/nestedhybrid.py
```diff
@@ -0,0 +1,542 @@
...
+    grid: xtgeo.Grid,
+    region: xtgeo.GridProperty,
+    region_id: int,
+) -> tuple[xtgeo.Grid, tuple[int, int, int]]:
+    """Crop *grid* to the bounding box of *region_id*.
+
+    Returns (cropped_grid, crop_origin) where *crop_origin* is the 0-based
+    (i, j, k) offset of the crop start inside the original grid.
+    """
+    region_values = region.values
```
alifbe, line 48: Should we check that region property has the same shape as grid?

src/fmu/tools/nestedhybridgrid/nestedhybrid.py
alifbe, line 1: One small improvement could be to create a small `dataclass` ijk to replace tuple[int, int, int]

src/fmu/tools/nestedhybridgrid/nestedhybrid.py
```diff
@@ -0,0 +1,542 @@
...
+            ref_js = range(cj * rrow, cj * rrow + rrow)
+            ref_ks = [ck * rlay]  # first k-layer (K- face)
+        elif face_dir == "k+":
+            # Target at lower K than mother → mother's K- face
+            direction = "K-"
+            ref_is = range(ci * rcol, ci * rcol + rcol)
+            ref_js = range(cj * rrow, cj * rrow + rrow)
+            ref_ks = [ck * rlay + rlay - 1]  # last k-layer (K+ face)
+        else:
+            continue
```
alifbe: should we trigger a warning instead of silently ignoring it?

src/fmu/tools/nestedhybridgrid/nestedhybrid.py
```diff
@@ -0,0 +1,542 @@
...
+
+    _logger.info(
+        "NNC table: %d cell pairs from %d boundary faces", len(rows), len(faces)
+    )
+    return pd.DataFrame(rows, columns=["I1", "J1", "K1", "I2", "J2", "K2", "DIRECTION"])
+
+
+def _set_actnum_by_region(
+    grid: xtgeo.Grid,
+    region_prop: xtgeo.GridProperty,
```
alifbe, line 253: Probably also check that the dimension of grid and property are identical

alifbe (approved):
Looks good to me 👍

Have a few suggestions for you to consider. But none of them are critical and we can always revisit it at later time.

src/fmu/tools/nestedhybridgrid/nestedhybrid.py
jcrivenaes: Not convinced, so I leave as is

src/fmu/tools/nestedhybridgrid/nestedhybrid.py
jcrivenaes: Already done I think before getting to this point, so leave as is.

PR #431, opened 2026-05-15
ENH: nested grid clean up

lahus-0:
Adjusted nested hybrid gridding workflow to ensure that layers in grid are sorted based on input order. This ensure resulting grid is consistent with standard geomodelling requirements (e.g. zonation, ....)

Cleaned up NEST_ID parameter as any xtgeo grid properties can be attached to grid and preserved. The NEST_ID is thus a copy of refine region and unnecessary.

## Checklist

- [X] Tests added (if not, comment why)
- [X] Test coverage equal or up from main (run pytest with `--cov=src/ --cov-report term-missing`)
- [?] If not squash merging, every commit passes tests
- [X] Appropriate [commit prefix](https://upgraded-funicular-eywe4gy.pages.github.io/developing/#commit-prefixes) and precise commit message used
- [X] All debug prints and unnecessary comments removed
- [X] Docstrings are correct and updated
- [X] Documentation is updated, if necessary
- [X] Latest main rebased/merged into branch
- [X] Added comments on this PR where appropriate to help reviewers
- [X] Moved issue status on project board

### DIFF
--- a/docs/nestedhybridgrid.rst
+++ b/docs/nestedhybridgrid.rst
@@ -41,6 +41,10 @@ The example here runs within RMS, but similar workflows can be created for file 
     # Load grid and region property
     grid = xtgeo.grid_from_roxar(project, "Simgrid")
     region = xtgeo.gridproperty_from_roxar(project, "Simgrid", "REGION")
+    
+    # Optionally load any other parameter you wish to preserve on grid
+    #zone = xtgeo.gridproperty_from_roxar(project, "Simgrid", "Zone")
+    #grid.append_prop(zone)
 
     # Create nested hybrid grid (e.g. refine region 2 by 2×2×1)
     merged, nnc_table = create_nested_hybrid_grid(
@@ -50,11 +54,12 @@ The example here runs within RMS, but similar workflows can be created for file 
     # store merged grid in RMS (or file)
     merged.to_roxar(project, "NestedHybrid")
 
-    merged.get_prop_by_name("NEST_ID")
-    nest_id.to_roxar(project, "NestedHybrid", nest_id.name)
+    # Optionally extract and store any necessary parameters from grid
+    region2 = merged.get_prop_by_name("REGION")
+    region2.to_roxar(project, "NestedHybrid", region2.name)
 
     # write the NNC pandas to disk; this will be applied for computing NNC's in the next script
-    nnc_table.write_csv("path_to_some_csv_file.csv", index=False)
+    nnc_table.to_csv("path_to_some_csv_file.csv", index=False)
 
 
 The next step is to do a rescaling from the original geogrid to the merged grid
@@ -64,6 +69,7 @@ Further, we need to create NNC transmissibilities and generate file for flow sim
 
 .. code-block:: python
 
+    import pandas as pd
     import xtgeo
     from fmu.tools.nestedhybridgrid import (
         nnc_to_flowsimulator_input,
@@ -74,7 +80,6 @@ Further, we need to create NNC transmissibilities and generate file for flow sim
 
     # Load grid and region property which may be stored in RMS
     nested = xtgeo.grid_from_roxar(project, GNAME)
-    region = nested.get_prop_by_name("NEST_ID")  # if needed for QC
 
     # load the NNC table
     nnc_table = pd.read_csv("path_to_some_nnc_file.csv")
@@ -88,7 +93,7 @@ Further, we need to create NNC transmissibilities and generate file for flow sim
 
     # compute transmissibilities. Note that flow simulators do this for the normal cells/faults
     # so strictly speaking, only nnc_hybrid is needed here.
-    tranx, trany, tranz, nnc_fault, nnc_hybrid, rbnd = merged.get_transmissibilities(
+    tranx, trany, tranz, nnc_fault, nnc_hybrid, rbnd = nested.get_transmissibilities(
         permx, permy, permz, ntg, nnc_table=nnc_table
     )
 
@@ -96,7 +101,7 @@ Further, we need to create NNC transmissibilities and generate file for flow sim
     nnc_to_flowsimulator_input(nnc_hybrid, "some_path/NNC_HYBRID.INC")
 
     # Or map NNCs onto grid properties for visualisation
-    tx_nnc, ty_nnc, tz_nnc = nnc_to_gridproperty(merged, nnc_hybrid)
+    tx_nnc, ty_nnc, tz_nnc = nnc_to_gridproperty(nested, nnc_hybrid)
     tx_nnc.to_roxar(project, GNAME, "TRANX_NNC_QC")  # etc
 
 Concepts
@@ -126,15 +131,6 @@ This table is passed to :meth:`xtgeo.Grid.get_transmissibilities` via the
 ``nnc_table`` parameter.  The transmissibility computation uses geometric
 face-overlap calculations (Sutherland–Hodgman algorithm) and two-point flux
 approximation (TPFA).
-
-NEST_ID property
-^^^^^^^^^^^^^^^^
-
-The merged grid carries a discrete property called ``NEST_ID``:
-
-- **0** — inactive hole cells (carved out to make room for the refined region)
-- **1** — mother (coarse) cells
-- **2** — refined cells
 
 Eclipse / OPM Flow export
 ^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/src/fmu/tools/nestedhybridgrid/nestedhybrid.py
+++ b/src/fmu/tools/nestedhybridgrid/nestedhybrid.py
@@ -130,6 +130,8 @@ def _compute_nnc_table(
     crop_origin: tuple[int, int, int],
     refinement: tuple[int, int, int],
     coarse_ncol: int,
+    lmap1: np.ndarray,
+    lmap2: np.ndarray,
 ) -> pd.DataFrame:
     """Compute NNC cell-pair mapping between mother and refined cells.
 
@@ -155,6 +157,9 @@ def _compute_nnc_table(
         crop_origin: 0-based ``(i0, j0, k0)`` origin of the crop box.
         refinement: ``(rcol, rrow, rlay)`` refinement factors.
         coarse_ncol: Number of columns in the coarse grid (grid1 in the merge).
+        lmap1: Numpy array with layer_mapping (input k -> output k) for grid1
+        lmap2: Numpy array with layer_mapping (input k -> output k) for grid2
+
 
     Returns:
         A DataFrame with columns ``I1, J1, K1, I2, J2, K2, DIRECTION``.
@@ -163,6 +168,7 @@ def _compute_nnc_table(
 
     rcol, rrow, rlay = refinement
     i0, j0, k0 = crop_origin
+
     # In the merged grid, grid2 (refined) starts after a 1-column gap:
     i_offset = coarse_ncol + 1
 
@@ -234,10 +240,10 @@ def _compute_nnc_table(
                         {
                             "I1": mi + 1,
                             "J1": mj + 1,
-                            "K1": mk + 1,
+                            "K1": lmap1[mk] + 1,
                             "I2": ri + i_offset + 1,
                             "J2": rj + 1,
-                            "K2": rk + 1,
+                            "K2": lmap2[rk] + 1,
                             "DIRECTION": direction,
                         }
                     )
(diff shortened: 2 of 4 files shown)


src/fmu/tools/nestedhybridgrid/nestedhybrid.py
```diff
@@ -365,30 +499,18 @@ def create_nested_hybrid_grid(
...
-        values=np.where(refined.get_actnum().values == 1, 2, 0).astype(np.int32),
-        codes={0: "inactive", 1: "mother", 2: "refined"},
-    )
-    refined.append_prop(nest_id_refined)
-
-    # 7. Merge the two grids.
-    merged = xtgeo.grid_merge(grid, refined)
+    # 8. Merge the two grids.
+    merged = xtgeo.grid_merge(grid, refined, olay, rlay)
     _logger.info("Merged grid dimensions: %s", merged.dimensions)
```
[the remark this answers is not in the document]
lahus-0: NEST_ID redundant hence removed

src/fmu/tools/nestedhybridgrid/nestedhybrid.py
```diff
@@ -271,6 +285,196 @@ def _set_actnum_by_region(
...
+        raise ValueError("Invalid input upscaling relationships")
+
+    ivo = iv.copy()
+    jvo = jv.copy()
+    kvo = kv.copy()
+
+    # mask the cells which are excluded from upscaling as value was 0
+    iv = np.ma.masked_where((iv == -1) | (jv == -1) | (kv == -1), iv)
+    jv = np.ma.masked_where((iv == -1) | (jv == -1) | (kv == -1), jv)
+    kv = np.ma.masked_where((iv == -1) | (jv == -1) | (kv == -1), kv)
```
alifbe: For consistency, perhaps compute the filter before overwriting iv with masked array.

```suggestion
    exclude_mask = (iv == -1) | (jv == -1) | (kv == -1)
    iv = np.ma.masked_where(exclude_mask, iv)
    jv = np.ma.masked_where(exclude_mask, jv)
    kv = np.ma.masked_where(exclude_mask, kv)
```

alifbe: Also, is it intentionally to modify iv, jv, and kv in this function? I'd recommend to create a new numpy array instead.

src/fmu/tools/nestedhybridgrid/nestedhybrid.py
```diff
@@ -271,6 +285,196 @@ def _set_actnum_by_region(
...
+
+    # shift region from input grid to geogrid to know which cells are refined or not
+    imap, jmap, kmap = up
+    iref, jref, kref = ref
+
+    iv = imap.values.reshape(-1) - 1
+    jv = jmap.values.reshape(-1) - 1
+    kv = kmap.values.reshape(-1) - 1
+    irv = iref.values
+    jrv = jref.values
```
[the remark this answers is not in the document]
lahus-0: delibrately updated in place in case same process is looped multiple times on top of each other. can do other way if others prefer.

alifbe (commented):
I suggest to bump up version requirement for xtgeo to at least 4.22.2 to fix testing issue.

https://github.com/equinor/fmu-tools/blob/3f6a4adc56489cc9eedd074d32f9e7bb6a9d3ff0/pyproject.toml#L47

src/fmu/tools/nestedhybridgrid/nestedhybrid.py
```diff
@@ -271,6 +285,196 @@ def _set_actnum_by_region(
...
+
+                lmap2 = lmap2.reshape(-1)
+                kvo[np.isin(cn, cnr)] = lmap2[np.isin(cl2, cnr)]
+
+                cmap2 = cmap2.reshape(-1)
+                ivo[np.isin(cn, cnr)] = cmap2[np.isin(cl2, cnr)]
+
+                rmap2 = rmap2.reshape(-1)
+                jvo[np.isin(cn, cnr)] = rmap2[np.isin(cl2, cnr)]
+
```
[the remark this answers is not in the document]
lahus-0: these can be large and intensive, but I don't see an easier what considering that the refinement levels can vary in ijk and hence the structure required becomes complex to build in general

lahus-0:
> I suggest to bump up version requirement for xtgeo to at least 4.22.2 to fix testing issue.
> 
> https://github.com/equinor/fmu-tools/blob/3f6a4adc56489cc9eedd074d32f9e7bb6a9d3ff0/pyproject.toml#L47

that worked

src/fmu/tools/nestedhybridgrid/nestedhybrid.py
```diff
@@ -271,6 +285,197 @@ def _set_actnum_by_region(
...
+                    jl2.reshape(-1) * (kv.max() + 1) + kl2.reshape(-1)
+                )
+
+                lmap2 = np.arange(dk, dtype=np.float32) + ok
+                lmap2 = np.tile(lmap2, di * dj).reshape((di, dj, dk))
+                lmap2 = np.repeat(lmap2, int(uri / ri), axis=0)
+                lmap2 = np.repeat(lmap2, int(urj / rj), axis=1)
+                lmap2 = np.repeat(lmap2, int(urk / rk), axis=2)
+
+                cmap2 = np.arange(di, dtype=np.float32) + iv.max() + 2
```
GibranAlfa: The upscaling remap seems to calculate the refined-grid position from the largest value found in the mapping (`iv.max()` etc.). That works only when the mapping includes the whole input grid. If some valid cells are excluded with `0`, the largest value can be too small, so refined cells may be mapped to the wrong place, even into the merge gap. I think this should use the actual input grid dimensions (`iref.ncol`, etc.) instead. A test with `0` exclusions should catch this.

docs/nestedhybridgrid.rst
GibranAlfa: Valid copilot comment but wrong placement, it should be here. The example uses `nnc_to_gridproperty(merged, nnc_hybrid)`, but that script defines the grid as `nested`, so it should use `nested` there.

GibranAlfa (commented):
I think the coverage of the tests should be improved. Right now it's only covering the happy path, it would be better if it also cover the edge cases and error cases.

src/fmu/tools/nestedhybridgrid/nestedhybrid.py
lahus-0: fair enough I'll change and adapt these when i get the chance

lahus-0:
@alifbe as discussed upscaling functionality removed to do these changes split into 2 phases.

alifbe:
> @alifbe as discussed upscaling functionality removed to do these changes split into 2 phases.

@lahus-0 👍Please update the PR title and description as well?

@tnatt @GibranAlfa Can you please review the PR again now that the mapping part is move to another PR? Thanks 🙏

src/fmu/tools/nestedhybridgrid/nestedhybrid.py
```diff
@@ -163,6 +167,16 @@ def _compute_nnc_table(
...
+
+    # generate the input to output layer mappings
+    lmap1 = np.arange(coarse_nlay, dtype=np.int32)
+    lmap1 = lmap1 + np.where(
+        lmap1 < k0,
+        0,
+        (rlay - 1) * np.minimum(int(refined_nlay / rlay), lmap1 - k0),
+    )
+    lmap2 = np.arange(refined_nlay, dtype=np.int32) + k0
+
```
tnatt: It seems to me that this relies on that the functionality for merging grids in xtgeo doesn't change the logic of how it creates layers in the merged grid. If so, this is quite fragile and can lead to incorrect nnc's if that happens. https://github.com/equinor/xtgeo/blob/302f399aacc8fe980da936fb479aaed48ec8308a/src/xtgeo/grid3d/_grid_merge.py#L63-L77

Would it be possible to separate out this logic in a function in xtgeo, which could be used in both packages? 
Or create these layer mappings up front here, and pass those as argument into `grid_merge` instead of the     `layer_offset` and `layer_refinement` ?

tnatt (commented):
Nice cleanup of the `NEST_ID` properties🙂 
I don't see any tests for the added functionality for mapping layers, could you add that?

src/fmu/tools/nestedhybridgrid/nestedhybrid.py
lahus-0: fmu tools here controls the process and calls the xtgeo functionality. The xtgeo is deliberately left more general to be usable for other purposes too. It would be possible to adjust optional xtgeo inputs to expect the lmap1 and lmap2 numpy arrays instead of the layer_offset and layer_refinement int. This would make the xtgeo inputs a bit more complex but would force alignment always with what is happening here. Let me know if you want that approach and I'll start sorting that, just note it will need updates to xtgeo first.

lahus-0: xtgeo fix sent (https://github.com/equinor/xtgeo/pull/1646). This now builds lmap1 and lmap2 in fmu.tools then sends them to xtgeo. Updating this here now but everything will fail until xtgeo merged.

lahus-0:
needs updated xtgeo with https://github.com/equinor/xtgeo/pull/1646 to work again. tests pass locally using this.

docs/nestedhybridgrid.rst
```diff
@@ -74,7 +80,7 @@ Further, we need to create NNC transmissibilities and generate file for flow sim
 
     # Load grid and region property which may be stored in RMS
     nested = xtgeo.grid_from_roxar(project, GNAME)
-    region = nested.get_prop_by_name("NEST_ID")  # if needed for QC
+    region = xtgeo.gridproperty_from_roxar(project, GNAME, "REGION")  # if needed for QC
```
alifbe: I guess this line is not needed for example?

tnatt:
Hi @lahus-0 :) I just made a new `4.23.0` release of xtgeo, could you update the pin in `pyproject.toml` to see if the tests will pass then?

lahus-0:
updated and all passed just let me know if I need to clean up more here now. I'll have a closer look and clean anything I see up when I work on the upscaling linkages afterwards.

src/fmu/tools/nestedhybridgrid/nestedhybrid.py
```diff
@@ -343,18 +361,21 @@ def create_nested_hybrid_grid(
...
-    nnc_table = _compute_nnc_table(
+    nnc_table, lmap1, lmap2 = _compute_nnc_table(
         region_prop=region,
         target_region_id=target_region_id,
         crop_origin=crop_origin,
         refinement=refinement,
         coarse_ncol=grid.ncol,
+        coarse_nlay=grid.nlay,
+        refined_nlay=refined.nlay,
     )
```
tnatt: I don't see any particular reason why we pass in `coarse_nlay` and `refined_nlay` into the `_compute_nnc_table` function, it seems it is only there to generate the layer mappings. 
A better pattern would be to have a separate function to generate the `lmap1` and `lmap2` outside the function, which is passed in instead.

This will make testing the logic for generating the layer mapping easier since it doesn't have to go through  `_compute_nnc_table`. Then `_compute_nnc_table` only returns the `nnc_table` which is more the expected behaviour from this function.

Something like the below 

```suggestion
    lmap1, lmap2 = _generate_layer_mappings(grid.nlay, refined.nlay, olay, rlay)
    
    nnc_table = _compute_nnc_table(
        region_prop=region,
        target_region_id=target_region_id,
        crop_origin=crop_origin,
        refinement=refinement,
        coarse_ncol=grid.ncol,
        lmap1=lmap1,
        lmap2 =lmap2 ,
)
```

lahus-0: done

src/fmu/tools/nestedhybridgrid/nestedhybrid.py
```diff
@@ -313,12 +350,11 @@ def create_nested_hybrid_grid(
         region: A :class:`xtgeo.GridProperty` whose values identify the
             regions (e.g. an integer region parameter).
         target_region_id: The region value to refine.
-        refinement: ``(ncol, nrow, nlay)`` refinement factors.
```
tnatt, line 316: Was this removed by mistake?

tnatt (approved):
This looks in general good 👍 
I would like to see more tests for the layer mapping produced by some more combinations of refinement and origins though. If you have time please add it to this PR, if not add it in a separate PR. We aim to make a new release of fmu-tools tomorrow before lunch.

src/fmu/tools/nestedhybridgrid/nestedhybrid.py
lahus-0: accidentally removed when i cut all the upscaling stuff out of this. fixed and more tests added

tnatt:
This looks good! Would you squash your commits and rebase the branch to main so that we can merge it @lahus-0?

lahus-0:
sure i will squash then rebase it all now. Thanks

lahus-0:
should be done now. been a while took me a bit longer than I remembered to rebase properly.

PR #438, opened 2026-06-05
ENH: nested grid preserves zonation

lahus-0:
preserve zonation from input grid when generating a nested hybrid grid

## Checklist

- [X] Tests added (if not, comment why)
- [X] Test coverage equal or up from main (run pytest with `--cov=src/ --cov-report term-missing`)
- [X] If not squash merging, every commit passes tests
- [X] Appropriate [commit prefix](https://upgraded-funicular-eywe4gy.pages.github.io/developing/#commit-prefixes) and precise commit message used
- [X] All debug prints and unnecessary comments removed
- [X] Docstrings are correct and updated
- [X] Documentation is updated, if necessary
- [X] Latest main rebased/merged into branch
- [X] Added comments on this PR where appropriate to help reviewers
- [?] Moved issue status on project board

### DIFF
--- a/src/fmu/tools/nestedhybridgrid/nestedhybrid.py
+++ b/src/fmu/tools/nestedhybridgrid/nestedhybrid.py
@@ -310,6 +310,28 @@ def _generate_layer_mappings(
     return (lmap1, lmap2)
 
 
+def _set_zonation(subgrid: dict, lmap: np.ndarray, nlay: int) -> dict:
+    """Create an updated subgrid dictionary for the merged grid.
+    Args:
+        subgrid: subgrid dictionary from input grid
+        lmap: layer mapping for grid1 - unrefined input grid
+        nlay: total number of layers in merged grid
+    """
+
+    updated_subgrid = {}
+
+    # sorted list of zones to add
+    zl = sorted(subgrid, key=lambda x: subgrid[x][0])
+
+    for zi in range(len(zl)):
+        zn = zl[zi]
+        zmin = lmap[subgrid[zn][0] - 1] + 1
+        zmax = nlay + 1 if zi == len(zl) - 1 else lmap[subgrid[zl[zi + 1]][0] - 1] + 1
+        updated_subgrid[zn] = range(zmin, zmax)
+
+    return updated_subgrid
+
+
 # ---------------------------------------------------------------------------
 # Public API
 # ---------------------------------------------------------------------------
@@ -377,6 +399,9 @@ def create_nested_hybrid_grid(
     grid = grid.copy()
     region = region.copy()
 
+    # Save input zonation
+    subgrid = grid.subgrids
+
     # Attach the region property to the grid.
     grid.append_prop(region)
 
@@ -422,6 +447,8 @@ def create_nested_hybrid_grid(
     # 7. Merge the two grids.
     merged = xtgeo.grid_merge(grid, refined, lmap1, lmap2)
     _logger.info("Merged grid dimensions: %s", merged.dimensions)
+    if subgrid is not None:
+        merged.subgrids = _set_zonation(subgrid=subgrid, lmap=lmap1, nlay=merged.nlay)
 
     return merged, nnc_table
 
--- a/tests/nestedhybridgrid/test_nestedhybrid.py
+++ b/tests/nestedhybridgrid/test_nestedhybrid.py
@@ -198,6 +198,37 @@ class TestCreateNestedHybridGrid:
         assert np.array_equal(lmap1, np.array([0, 1, 11]))
         assert np.array_equal(lmap2, np.arange(20) + 1)
 
+    def test_zonation(self):
+        """Input grid with zonation returns a valid zonation"""
+        grid, region, rid = _make_box_grid_with_region(dimension=(6, 6, 2))
+        grid.subgrids = {"ZONE1": [1], "ZONE2": [2]}
+        merged, nnc_table = create_nested_hybrid_grid(
+            grid, region, rid, refinement=(2, 2, 2)
+        )
+        subgrids_nlay = merged.get_subgrids()
+        assert subgrids_nlay["ZONE1"] == 2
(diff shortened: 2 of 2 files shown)


src/fmu/tools/nestedhybridgrid/nestedhybrid.py
```diff
@@ -309,6 +309,28 @@ def _generate_layer_mappings(
...
+    updated_subgrid = {}
+
+    # sorted list of zones to add
+    zl = sorted(subgrid, key=lambda x: subgrid[x][0])
+
+    for zi in range(len(zl)):
+        zn = zl[zi]
+        zmin = lmap[subgrid[zn][0] - 1] + 1
+        zmax = nlay + 1 if zi == len(zl) - 1 else lmap[subgrid[zl[zi + 1]][0] - 1] + 1
+        updated_subgrid[zn] = range(zmin, zmax)
```
alifbe, line 330:
```suggestion
    for zone_name, zone_range  in subgrid.items():
        zmin = lmap[min(zone_range) - 1] + 1
        zmax = lmap[max(zone_range) - 1] + 1
        updated_subgrid[zone_name] = range(zmin, zmax)
```

alifbe: I think `subgrid` will always be sorted dowward so we don't really need to do sorting in advance.

src/fmu/tools/nestedhybridgrid/nestedhybrid.py
```diff
@@ -309,6 +309,28 @@ def _generate_layer_mappings(
     return (lmap1, lmap2)
 
 
+def _set_zonation(subgrid: dict, lmap: np.ndarray, nlay: int) -> dict:
```
alifbe, line 313: We don't need `nlay` anymore as it can be deduced from `lmap`.

```suggestion
def _set_zonation(subgrid: dict, lmap: np.ndarray) -> dict:
```

alifbe (commented):
Some refactoring to be consider 😄

tests/nestedhybridgrid/test_nestedhybrid.py
```diff
@@ -198,6 +198,18 @@ def test_lmap_generation_ref10(self):
...
+        """Input grid with zonation returns a valid zonation"""
+        grid, region, rid = _make_box_grid_with_region(dimension=(6, 6, 2))
+        grid.subgrids = {"ZONE1": [1], "ZONE2": [2]}
+        merged, nnc_table = create_nested_hybrid_grid(
+            grid, region, rid, refinement=(2, 2, 2)
+        )
+        subgrids = merged.get_subgrids()
+
+        assert subgrids["ZONE1"] == 2
+        assert subgrids["ZONE2"] == 2
```
tnatt: Now we only test that they have the correct number of layers. I think we should also ensure that the layer ranges are correct. You could do this with the `merged.subgrids` property 🙂 

```suggestion
        subgrids_nlay = merged.get_subgrids()       
        assert subgrids_nlay["ZONE1"] == 2
        assert subgrids_nlay["ZONE2"] == 2
               
        assert merged.subgrids["ZONE1"] == range(1, 3)
        assert merged.subgrids["ZONE2"] == range(3, 5)
```

tests/nestedhybridgrid/test_nestedhybrid.py
```diff
@@ -198,6 +198,18 @@ def test_lmap_generation_ref10(self):
...
+
```
tnatt, line 214: I think we should also add a test where the region starts with a offset in layer.

    def test_zonation_with_layer_offset(self):
        """Input grid with zonation returns a valid zonation"""
        grid, region, rid = _make_box_grid_with_region(dimension=(6, 6, 3))
        grid.subgrids = {"ZONE1": [1], "ZONE2": [3]}

        region.values[:,:,0] = 1 # set value in first layer to not be rid

        ... rest of test

src/fmu/tools/nestedhybridgrid/nestedhybrid.py
lahus-0: no it can't. this is lmap for the coarse grid only and if the fine grid extends through entire grid vertically there will be more layers than the maximum value in lmap

tests/nestedhybridgrid/test_nestedhybrid.py
lahus-0: i will add that.

src/fmu/tools/nestedhybridgrid/nestedhybrid.py
lahus-0: if you are sure of that i am happy to change it. I went with the assumption better safe than sorry and that dictionaries in general aren't guaranteed to be sorted.

lahus-0: actually no in general here we need the list anyway. lmap here only shows the coarse grid layers and refinement will add extra vertical layers in places. You cannot just assume range extends to coarse grid lmap +1. 2 layers vertically on input a vertical refinement of 2 in some lateral only area. in this case

lmap=[0,2] but total grid has 4 layers. This is why we need the total number of layers as an input and need to look at the start of the next zone instead of end of current zone.

lahus-0:
Resolved the comments i agreed with. The others i don't think will work. i can squash for merge if needed.

lahus-0:
It's been about a week. Any more comments on what needs fixed here or not? I have added more tests which I believe cover all the suggestions. For the other suggested changes I have explained why it won't work.

alifbe:
> It's been about a week. Any more comments on what needs fixed here or not? I have added more tests which I believe cover all the suggestions. For the other suggested changes I have explained why it won't work.

Sorry, we were in conference this week. But I think the PR looks good to me. What do you think @tnatt ?

tnatt (approved):
Yes fine by me as well!

lahus-0:
I'll squash and rebase now then

Issue #446, opened 2026-06-15
Flaws in domainconversion _extend_incube_create_speed_cube_average


anbec12:
Issue 1: When creating the speedcube, the actual speedcube values are inter- and extrapolated with np.interp, which used constant extrapolation on the right side, meaning the maximum depth value of the input. This leads to wrong velocities below the lowermost horizon used as input. Better to inter- and extrapolate in time-depth domain which calculates correct velocities.

Issue 2: The code iterates over each trace for interpolation and takes significant time for larger cubes. Try to optimize speed.

The two issues are reported as one because the solution is going to be in one pull request (solution has already been coded before issue has been made).

PR #448, opened 2026-06-15
BUG: Fix stretch in converted seismic when MSL is given and avoid flo…

### LINKED ISSUE #445: Seismic gets stretched above first subsurface horizon when using MSL as input

anbec12:
…ating-point errors

Resolves #445 and #446 

Added several fixes that were related. It became therefore a relatively large pull-request.

1. In the tests, the issue with the commented out line with parameters [(None, None, None), (1.1, 0, 82.5)] for test_proposing_zinc_etc_depthconvert was resolved by adding correct support for None as input in dconvert.py _derive_result_cube_design and by dealing with potential floating-point errors when calculating the nlay parameter.

2. There was a substantial methodological issue with the way the velocity or slowness cubes were generated in dconvert.py _extend_incube_create_speed_cube_average. Originally, the average velocity or slowness maps were interpolated and extrapolated with np.interp to get the upsampled 'speed' function corresponding to the trace sampling. However, this could lead to stretching of the converted seismic in the overburden and wrong velocities below the last horizon. I have now changed the interpolation to linear interpolation of the time-depth function adding linear right-slope extrapolation which resolves the issues. This required also an update of _resample_surfaces_to_cube, a helper function which is only called by _extend_incube_create_speed_cube_average, returning time and depth surfaces instead of time/depth and velocity/slowness surfaces, depending if converting time->depth or vice versa.

3. When providing MSL horizons for time and depth (MSL: everywhere z = 0), one got different results compared to the case they were not provided. The results should have been identical (now included tests for this). The result with the MSL given showed stretched seismic in the overburden. This issue was solved by case differentiation if MSL was input in dconvert.py _velo_maps_average and likewise _slow_maps_average.

The features were related because they failed on the test_depth_cube_values_with_msl and test_depth_cube_values, depending on the input. I have also added a small sine-wave cube for testing, because I think the comparison before-after is more valid when you have data throughout the cube. The cube will also be used in a future PR with respect to amplitude preservation and trace interpolation.

## Checklist

- [ ] Appropriate [commit prefix](https://upgraded-funicular-eywe4gy.pages.github.io/developing/#commit-prefixes) and precise commit message used

### DIFF
--- a/src/fmu/tools/domainconversion/dconvert.py
+++ b/src/fmu/tools/domainconversion/dconvert.py
@@ -245,6 +245,11 @@ class DomainConversion:
             vspeed.values *= 2000
             vel.append(vspeed)
 
+        d1_values = self._d_surfaces[1].values
+        t1_values = self._t_surfaces[1].values
+        if np.allclose(d1_values, 0.0) and np.allclose(t1_values, 0.0):
+            vel[0] = vel[1]
+
         vel.insert(0, vel[0])
         self._v_surfaces = vel
         _logger.debug("Create velocity maps for average velocities from MSL...  DONE")
@@ -266,6 +271,11 @@ class DomainConversion:
             vslow.values = np.divide((t1.values - t0.values), ddiff)
             vslow.values /= 2000
             slow.append(vslow)
+
+        d1_values = self._d_surfaces[1].values
+        t1_values = self._t_surfaces[1].values
+        if np.allclose(d1_values, 0.0) and np.allclose(t1_values, 0.0):
+            slow[0] = slow[1]
 
         slow.insert(0, slow[0])
         self._s_surfaces = slow
@@ -318,23 +328,23 @@ class DomainConversion:
     ) -> tuple[list[xtgeo.RegularSurface], list[xtgeo.RegularSurface]]:
         """Resample surfaces to match the input cube.
 
-        The time and velocity surfaces do already exist, but they are not sampled the
+        The time and depth surfaces do already exist, but they are not sampled the
         actual cube, so we do that here.
         """
         _logger.debug("Resample surfaces to cube")
-        xsurfs = self._t_surfaces if time2depth else self._d_surfaces  # time or depth
-        ssurfs = self._v_surfaces if time2depth else self._s_surfaces  # "speed"
+        xsurfs = self._t_surfaces if time2depth else self._d_surfaces
+        ysurfs = self._d_surfaces if time2depth else self._t_surfaces
 
         incube_template = xtgeo.surface_from_cube(incube, value=0)
 
         x_surfaces = self._resample_check_surfaces(
             xsurfs, incube_template, fill=True, ensure_consistency=True
         )
-        s_surfaces = self._resample_check_surfaces(
-            ssurfs, incube_template, fill=True, ensure_consistency=False
-        )  # intended: do not ensure consistency of velocities
+        y_surfaces = self._resample_check_surfaces(
+            ysurfs, incube_template, fill=True, ensure_consistency=True
+        )
 
-        return x_surfaces, s_surfaces
+        return x_surfaces, y_surfaces
 
     @staticmethod
     def max_depth_for_cube(cube: xtgeo.Cube) -> float:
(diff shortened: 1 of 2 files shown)


src/fmu/tools/domainconversion/dconvert.py
```diff
@@ -368,24 +378,39 @@ def _extend_incube_create_speed_cube_average(
 
         speedcube_values = np.zeros_like(_incube.values)
```
tnatt: I know the code above was not touched in your PR, but could it be 😄 
The `_speed` and `_incube` properties are assigned multiple times. Also I don't see any purpose of having these as internal variables inside an internal function, could we rename them? 
```python
cube = self._recreate_cube_from_msl(incube, resample=True)
_logger.debug("Incube extended to MSL dimensions: %s", _incube.values.shape)

cube_arr = [cube.zori + n * cube.zinc for n in range(cube.values.shape[2])]

speedcube = cube.copy()
speedcube_values = np.zeros_like(cube.values)
```

src/fmu/tools/domainconversion/dconvert.py
```diff
@@ -368,24 +378,39 @@ def _extend_incube_create_speed_cube_average(
...
-        x_surfaces, s_surfaces = self._resample_surfaces_to_cube(_incube, time2depth)
+        # Resample surfaces to match the input cube
+        x_surfaces, y_surfaces = self._resample_surfaces_to_cube(_incube, time2depth)
 
         xlen = len(x_surfaces)
-        slen = len(s_surfaces)
+        ylen = len(y_surfaces)
 
-        assert xlen == slen
+        assert xlen == ylen
```
tnatt: We normally don't have asserts inside production code, it would be better to handle this with an exception and give a proper message to the developer/user.
```suggestion
        if len(x_surfaces) != len(y_surfaces):
            raise ValueError("x_surfaces and y_surfaces must have the same length")
```

src/fmu/tools/domainconversion/dconvert.py
```diff
@@ -368,24 +378,39 @@ def _extend_incube_create_speed_cube_average(
...
 
         for i in range(speedcube_values.shape[0]):
             for j in range(speedcube_values.shape[1]):
                 xmap = [x_surfaces[num].values[i, j] for num in range(xlen)]
-                smap = [s_surfaces[num].values[i, j] for num in range(slen)]
+                ymap = [y_surfaces[num].values[i, j] for num in range(ylen)]
```
tnatt: To be a bit more pythonic you could rewrite this to 👇 and remove the `xlen` and `ylen` assignments above 🙂 
```
xmap = [surf.values[i, j] for surf in x_surfaces]
ymap = [surf.values[i, j] for surf in y_surfaces]
```

src/fmu/tools/domainconversion/dconvert.py
```diff
@@ -368,24 +378,39 @@ def _extend_incube_create_speed_cube_average(
...
+
+                # Linear interpolation and extrapolation of time-depth function
+                y_arr = np.interp(incube_arr, xmap, ymap)
+                right_slope = (ymap[-1] - ymap[-2]) / (xmap[-1] - xmap[-2])
+                y_arr = np.where(
+                    incube_arr > xmap[-1],
+                    ymap[-1] + right_slope * (incube_arr - xmap[-1]),
+                    y_arr,
+                )
```
tnatt, line 392: This could be separated out into a function, and then you can write tests specifically for this function to ensure that the logic is as expected.

```suggestion
                # Linear interpolation and extrapolation of time-depth function
                y_arr = _interpolate_time_depth_function(xmap, ymap, incube_arr)
```

src/fmu/tools/domainconversion/dconvert.py
```diff
@@ -447,7 +476,14 @@ def _derive_result_cube_design(
...
         # assure that zori is a whole number multiple of zinc_depth starting from 0.0
         zori = nlay_above * zinc_actual
-        nlay = int((zmax_new - zori) / zinc_actual) + 1 + nlay_above
+
+        # avoid floating-point errors
+        n = (zmax_new - zori) / zinc_actual
+        if np.isclose(n, np.rint(n), atol=5e-4):
+            n = int(np.rint(n))
+        else:
+            n = int(np.floor(n))
```
tnatt, line 476: I wonder would `round()` do what you want here? 
I see the original code used `int` which just truncates the decimals and can make `2.999999` become `2`. `round()` would instead return `3` for this case.
```suggestion
       n = round((zmax_new - zori) / zinc_actual)
```

tnatt: Also the existing code has quite large functions, which makes it hard to set up good tests for smaller logic inside. This here for instance would need to be tested through checks against properties of the cubes produced.

Any chance some of this computing number of layers could be refactored into a smaller function, so that we could add test on it?

src/fmu/tools/domainconversion/dconvert.py
anbec12: That's true. The first assignments of `_incube` and `_speed` have no effect. Have tested it. I delete them.
I also agree with the naming suggestions. Have renamed the cubes in `_extend_incube_create_speed_cube_average`.
Checked the rest of the code as well for other weird cube names. Thus, renamed `xinv_cube` to `cube` in `_derive_result_cube_design`

src/fmu/tools/domainconversion/dconvert.py
anbec12: Agree

src/fmu/tools/domainconversion/dconvert.py
anbec12: Agree

src/fmu/tools/domainconversion/dconvert.py
anbec12: In a future PR, the two for-loops over i and j will be replaced by a more efficient calculation method with just one for-loop over k indices. Thus I don't want to separate out the inter- and extrapolation part in a separate function because it will be replaced anyway.

src/fmu/tools/domainconversion/dconvert.py
anbec12: There are actually two tests set up for this already: `test_proposing_zinc_etc_depthconvert` and `test_proposing_zinc_etc_timeconvert`. 
Both fail when I replace the lines with `n = round((zmax_new - zori) / zinc_actual)`. 
I agree with the input and expected output set up for these tests. Therefore, I would like to keep my suggestion, not sure if `atol ==5e-4` is a good suggestions though, but at least reasonable, I think.
The purpose is to round down and at the same time avoid floating-point errors. Rounding up with `round(..)` for e.g. `n = 99.6` is not what you want.

Regarding the second comment about refactoring, given that two tests are already set up I think it is covered already. Thus, I suggest to keep as is.

src/fmu/tools/domainconversion/dconvert.py
anbec12: Replaced also another assertion in `_domain_convert_cube` with `if ...: raise ValueError(...)`

src/fmu/tools/domainconversion/dconvert.py
tnatt: Then I agree 👍

src/fmu/tools/domainconversion/dconvert.py
tnatt: Ok, thanks for checking! I was not aware of the tests, then I agree lets keep it like this then 👍

src/fmu/tools/domainconversion/dconvert.py
tnatt: Super!

tnatt (approved):
Very nice work :) just fix the formatting, and then I think this is good to go!

PR #449, opened 2026-06-17
ENH: Added FFT trace interpolation for amplitude preservation

### LINKED ISSUE #447: The domain conversion is not amplitude preserving
When using the domainconversion to convert time seismic, output cubes in depth do not show 100% the same amplitudes as the input. This becomes especially apparent when doing a full round trip (time->depth->time) and a velocity model with constant velocity of 2000 m/s, because then the time and depth cubes should be identical.
The issue is related to using np.interp (linear interpolation) for trace interpolation to the new_vertical_axis in domainconversion _domain_convert_cube. An amplitude preserving trace interpolation method should be implemented here.

anbec12:
Resolves #447 

Added FFT method for trace interpolation to achieve amplitude preservation. 

The 'linear' method with np.interp, that was in the code before, is fast but does not preserve amplitudes. The degree of amplitude loss depends on sampling interval and frequency content of the input. The amplitude loss is particularly visible when running a round trip time->depth->time. 

A new input argument 'method' has been added and the new 'fft' method is set as default. The 'linear' method is kept as alternative that could be used when speed is  more important than accuracy.

## Checklist

### DIFF
--- a/src/fmu/tools/domainconversion/dconvert.py
+++ b/src/fmu/tools/domainconversion/dconvert.py
@@ -1,11 +1,14 @@
 from __future__ import annotations
 
 import logging
+import os
+import warnings
 from dataclasses import dataclass, field
-from typing import Generator
+from typing import Generator, Literal
 
 import numpy as np
 import xtgeo
+from scipy.fft import irfft, next_fast_len, rfft
 
 _logger = logging.getLogger(__name__)
 
@@ -695,6 +698,7 @@ class DomainConversion:
         zmax_proposed: float | None = None,
         undefined: float = -999.25,
         time2depth: bool = True,
+        method: Literal["fft", "linear"] = "linear",
     ) -> xtgeo.Cube:
         """Generic domain conversion of a cube.
 
@@ -721,15 +725,19 @@ class DomainConversion:
 
         delta = xcube.zinc / 2000 if time2depth else xcube.zinc * 2000
         vertical = xcube.copy()  # "vertical" is "times" if time2depth is True
-        varr = [0 + n * delta for n in range(xcube.values.shape[2])]
-        vertical.values[:, :, :] = varr
+        n = xcube.values.shape[2]
+        vertical.values[:, :, :] = np.arange(n, dtype=vertical.values.dtype) * delta
         result_values = scube.values * vertical.values
 
         rcube = self._depth_cube if time2depth else self._time_cube
 
         zmax_actual = self.max_depth_for_cube(rcube)
 
-        _logger.debug("Zmax for depth cube is %s", zmax_actual)
+        _logger.debug(
+            "Zmax for domain converted cube is %s (time2depth: %s)",
+            zmax_actual,
+            time2depth,
+        )
 
         start = rcube.zori
         stop = zmax_actual + rcube.zinc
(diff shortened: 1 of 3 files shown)


anbec12:
This PR uses the partly the same tests / test data as the previous PR #448. However, the code in dconvert.py is independent.

anbec12:
I have updated the docstring with the 'method' before I send the PR for review, but I guess the documentation page also needs to be updated. I don't know how to do that.

src/fmu/tools/domainconversion/dconvert.py
```diff
@@ -643,33 +646,145 @@ def _domain_convert_cube(
...
+            # with negligable amplitude loss.
+            zinc_fine_target = min(1.0, xcube.zinc / 2)
+
+            # Block_size is the number of traces processed in batch.
+            # Tested on linux server with 42 MB cache and 4 workers.
+            block_size = 512
+
+            # Half of available CPUs used for parallel computation
+            ncpu = os.cpu_count()
+            workers = max(1, ncpu) // 2 if ncpu else 1
```
tnatt: This will be a problems when using ERT and running on the lsf cluster since `os.cpu_count()` will return the total CPUs on the node, not those allocated to the realization through the [NUM_CPU](https://fmu-docs.equinor.com/docs/ert/reference/configuration/keywords.html#num-cpu) keyword in ERT.

This will steal CPU's from other users/jobs on the node. I discussed with Scout to find a solution. Turns out the environment variable `OMP_NUM_THREADS` is set based on the `NUM_CPU` [in fmu_site_configuration](https://github.com/equinor/fmu-site-configuration/blob/0edc61d306fe0a9e7dce40991fb3300e6c9e7fb3/fmu_site_configuration/site_config.py#L62)
This variable is a common environment variable used by e.g. `numpy` to control parallelism.

We could utilize this to optimize for both cluster runs and local runs. 


```suggestion
            # If running on the lsf cluster, use the number of CPUs allocated         
            if ncpu := os.environ.get("OMP_NUM_THREADS"):
                 workers = int(ncpu)
            else:
                 # If running locally use half of available CPUs
                 ncpu = os.cpu_count() or 1
                 workers = max(1, ncpu // 2)

```

src/fmu/tools/domainconversion/dconvert.py
```diff
@@ -742,6 +857,7 @@ def depth_convert_cube(
         zmin: float | None = None,
         zmax: float | None = None,
         undefined: float = -999.25,
+        method="fft",
```
tnatt: I know it might only be `sim2seis` that is using this `DomainConversion`, but there is no way to be sure that this method is not in use in some user scripts. Hence we should not change the default method without a deprecation period of the previous default `linear` method as that might interfere with users model results.

Suggest setting the default method to `linear` here and add a `FutureWarning` for users not using `fft` that that will be the default algoritm in the near future, and that they need to explicitly set `method="linear"` to keep the same results. Then `sim2seis` need to explicitly set `method="fft"` to get this improved algorithm before it is default.

src/fmu/tools/domainconversion/dconvert.py
```diff
@@ -775,6 +893,7 @@ def time_convert_cube(
         tmin: float | None = None,
         tmax: float | None = None,
         undefined: float = -999.25,
+        method="fft",
```
tnatt: Same comment as above

src/fmu/tools/domainconversion/dconvert.py
```diff
@@ -643,33 +646,145 @@ def _domain_convert_cube(
...
+            "Zmax for domain converted cube is %s (time2depth: %s)",
+            zmax_actual,
+            time2depth,
+        )
 
         start = rcube.zori
         stop = zmax_actual + rcube.zinc
         num_steps = int((stop - start) / (rcube.zinc - np.finfo(np.float32).eps))
         new_vertical_axis = np.linspace(start, zmax_actual, num_steps)
+        # new_vertical_axis = np.linspace(start, stop, num_steps)
```
tnatt: leftover

src/fmu/tools/domainconversion/dconvert.py
```diff
@@ -617,6 +619,7 @@ def _domain_convert_cube(
         zmax_proposed: float | None = None,
         undefined: float = -999.25,
         time2depth: bool = True,
+        method: str = "fft",
```
tnatt: You could improve the type hint here by using a `Literal`
```suggestion
        method: Literal["fft", "linear"] = "fft",
```

src/fmu/tools/domainconversion/dconvert.py
anbec12: Ok. Changed to default 'linear'.

src/fmu/tools/domainconversion/dconvert.py
anbec12: Here also updated

src/fmu/tools/domainconversion/dconvert.py
anbec12: Deleted the line

src/fmu/tools/domainconversion/dconvert.py
anbec12: Updated with your suggestion

src/fmu/tools/domainconversion/dconvert.py
anbec12: Added a `FutureWarning` to both `depth_convert_cube` and `time_conert_cube`

src/fmu/tools/domainconversion/dconvert.py
```diff
@@ -742,6 +861,7 @@ def depth_convert_cube(
         zmin: float | None = None,
         zmax: float | None = None,
         undefined: float = -999.25,
+        method="linear",
```
tnatt: Missing typehint here

src/fmu/tools/domainconversion/dconvert.py
```diff
@@ -775,6 +905,7 @@ def time_convert_cube(
         tmin: float | None = None,
         tmax: float | None = None,
         undefined: float = -999.25,
+        method="linear",
```
tnatt: Here as well

src/fmu/tools/domainconversion/dconvert.py
```diff
@@ -643,33 +647,148 @@ def _domain_convert_cube(
...
+            # Tested on linux server with 42 MB cache and 4 workers.
+            block_size = 512
+
+            # If running on the lsf cluster, use the number of CPUs allocated
+            if ncpu := os.environ.get("OMP_NUM_THREADS"):
+                workers = int(ncpu)
+            else:
+                # If running locally use half of available CPUs
+                ncpu = os.cpu_count() or 1
+                workers = max(1, ncpu // 2)
```
tnatt: my suggestion was lacking conversion to int here. This probably solves the linting issues
```suggestion
                ncpu = int(os.cpu_count() or 1)
                workers = max(1, ncpu // 2)
```

tnatt: no ignore this.. this conversion should not be needed since cpu_count returns an integer or None

tnatt: I'll quickly have a second look at why the CI is failing

tnatt: ok I see that `mypy` gets confused about the assignment of `ncpu` above.. hence it thinks `ncpu` is of type `str` or `None`.

Lets just rewrite it by making another variable name to make mypy happy 

            # If running on the lsf cluster, use the number of CPUs allocated         
            if omp_threads := os.environ.get("OMP_NUM_THREADS"):
                workers = int(omp_threads)
            else:
                 # If running locally use half of available CPUs
                 ncpu = os.cpu_count() or 1
                 workers = max(1, ncpu // 2)

tnatt:
> I have updated the docstring with the 'method' before I send the PR for review, but I guess the documentation page also needs to be updated. I don't know how to do that.

It should be enough to update this file with relevant info
https://github.com/equinor/fmu-tools/blob/main/docs/domain_conversion.rst

tnatt (approved):
LGTM, would you add the two last type hints as well ? 🙂

PR #452, opened 2026-06-18
ENH: Reuse speedcube when looping over several cubes

### LINKED ISSUE #451: Reuse cubes when domain converting several input cubes with the same velocity model
When you loop over several cubes for domain conversion, the speedcube is generated by interpolation from surfaces for each cube. This is redundant and can take significant time. 

Proposed enhancement: Reuse speedcube if geometries of the speedcube and input seismic cube are the same.

anbec12:
Resolves #451

Check if existing speedcube (self._vcube_t or self._scube_d) has same geometry as input seismic cube (self._incube). If yes, short-cut the function _extend_incube_create_speed_cube_average and avoid interpolating the speedcube again.

This saves substantial time when looping over several cubes that share the geometry, which is normal in Sim2seis.

I have added a test that checks the logger message for "Reuse..." as expected for the second cube to be converted. However, I am not sure whether the test is necessary and helpful. The code in dconvert.py has a check function _same_cube_geometry built in and the rest of the updates is very short. So not sure about the test design.

## Checklist

### DIFF
--- a/src/fmu/tools/domainconversion/dconvert.py
+++ b/src/fmu/tools/domainconversion/dconvert.py
@@ -351,6 +351,23 @@ class DomainConversion:
         """Calculate the maximum depth for a cube."""
         return cube.zori + (cube.nlay - 1) * cube.zinc
 
+    @staticmethod
+    def _same_cube_geometry(cube1: xtgeo.Cube, cube2: xtgeo.Cube) -> bool:
+        """Check if two cubes have the same geometry."""
+        keys = ("xori", "yori", "zori", "xinc", "yinc", "zinc", "rotation", "yflip")
+
+        if cube1.dimensions != cube2.dimensions:
+            return False
+
+        for k in keys:
+            atol = 0.01 if k in ["xori", "yori"] else 1e-06
+
+            v1, v2 = float(getattr(cube1, k)), float(getattr(cube2, k))
+            if not np.isclose(v1, v2, atol=atol):
+                return False
+
+        return True
+
     def _extend_incube_create_speed_cube_average(
         self, incube: xtgeo.Cube, time2depth: bool = True
     ) -> None:
@@ -362,6 +379,24 @@ class DomainConversion:
         _logger.debug("Incube dimensions: %s", incube.values.shape)
         cube = self._recreate_cube_from_msl(incube, resample=True)
         _logger.debug("Incube extended to MSL dimensions: %s", cube.values.shape)
+
+        # Check if we can reuse existing speedcube or need to create a new one
+        if time2depth:
+            if self._vcube_t is not None and self._same_cube_geometry(
+                cube, self._vcube_t
+            ):
+                _logger.debug("Reuse velocity cube for time->depth conversion")
+                _logger.debug("Update internal time cube")
+                self._time_cube = cube
+                return
+        else:
+            if self._scube_d is not None and self._same_cube_geometry(
+                cube, self._scube_d
+            ):
+                _logger.debug("Reuse slowness cube for depth->time conversion")
+                _logger.debug("Update internal depth cube")
+                self._depth_cube = cube
+                return
 
         result_description = "velocity" if time2depth else "slowness"
         _logger.debug("Create average %s cube", result_description)
--- a/tests/domainconversion/test_domainconversion.py
+++ b/tests/domainconversion/test_domainconversion.py
@@ -332,7 +332,7 @@ def test_depth_cube_values(smallsinecube: xtgeo.Cube) -> None:
 
     new_depth_cube = dc.depth_convert_cube(smallsinecube, zinc=1.0, zmin=0, zmax=100)
 
-    assert np.allclose(smallsinecube.values, new_depth_cube.values, atol=0.0001)
+    assert np.allclose(smallsinecube.values, new_depth_cube.values, atol=0.021)
 
 
 def test_depth_cube_values_with_msl(smallsinecube: xtgeo.Cube) -> None:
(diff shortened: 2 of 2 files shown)


src/fmu/tools/domainconversion/dconvert.py
```diff
@@ -341,6 +341,36 @@ def max_depth_for_cube(cube: xtgeo.Cube) -> float:
...
+            v1, v2 = float(getattr(cube1, k)), float(getattr(cube2, k))
+            if not np.isclose(v1, v2, atol=1e-6):
+                diffs[k] = (v1, v2)
+
+        for k in utm_float_keys:
+            v1, v2 = float(getattr(cube1, k)), float(getattr(cube2, k))
+            if not np.isclose(v1, v2, atol=1e-2):
+                diffs[k] = (v1, v2)
+
+        return not diffs
```
tnatt: Instead of building a dictionary of differences here, it could be more simply done by early returns. 
```suggestion
        for k in int_keys:
            v1, v2 = getattr(cube1, k), getattr(cube2, k)
            if v1 != v2:
                return False

        for k in float_keys:
            v1, v2 = float(getattr(cube1, k)), float(getattr(cube2, k))
            if not np.isclose(v1, v2, atol=1e-6):
                return False

        for k in utm_float_keys:
            v1, v2 = float(getattr(cube1, k)), float(getattr(cube2, k))
            if not np.isclose(v1, v2, atol=1e-2):
                return False

        return True
```

src/fmu/tools/domainconversion/dconvert.py
```diff
@@ -341,6 +341,36 @@ def max_depth_for_cube(cube: xtgeo.Cube) -> float:
...
+        diffs = {}
+
+        for k in int_keys:
+            v1, v2 = getattr(cube1, k), getattr(cube2, k)
+            if v1 != v2:
+                diffs[k] = (v1, v2)
+
+        for k in float_keys:
+            v1, v2 = float(getattr(cube1, k)), float(getattr(cube2, k))
+            if not np.isclose(v1, v2, atol=1e-6):
```
tnatt: This will raise `TypeError` if one of the attributes returned from `getattr` are `None`. Do you know if this could happen?

src/fmu/tools/domainconversion/dconvert.py
```diff
@@ -341,6 +341,36 @@ def max_depth_for_cube(cube: xtgeo.Cube) -> float:
...
+        """
+        int_keys = ("ncol", "nrow", "nlay", "yflip")
+        float_keys = ("zori", "xinc", "yinc", "zinc", "rotation")
+        utm_float_keys = ("xori", "yori")
+        diffs = {}
+
+        for k in int_keys:
+            v1, v2 = getattr(cube1, k), getattr(cube2, k)
+            if v1 != v2:
+                diffs[k] = (v1, v2)
```
tnatt: For readability maybe we could do this check on dimensions instead 🙂 ? And then you could remove the `int_keys` and check the `yflip` together with the other float keys? Integers should work well in the `np.isclose` also.

`keys = ("zori", "xinc", "yinc", "zinc", "rotation", "yflip")`
```suggestion
        if cube1.dimensions != cube2.dimensions:
            return False
```

src/fmu/tools/domainconversion/dconvert.py
```diff
@@ -359,6 +389,28 @@ def _extend_incube_create_speed_cube_average(
 
         _logger.debug("Incube extended to MSL dimensions: %s", _incube.values.shape)
 
+        # Check if we can reuse existing speedcube or need to create a new one
+        if time2depth:
+            if (
+                hasattr(self, "_vcube_t")
+                and bool(self._vcube_t)
+                and self._same_cube_geometry(_incube, self._vcube_t)
+            ):
```
tnatt: From what I can tell `_vcube_t` will always exist as an attribute in the dataclass, but may be `None` before it is populated. Hence it’s more direct and readable to check for `None` explicitly 🙂 
```suggestion
            if self._vcube_t is not None and self._same_cube_geometry(
                _incube, self._vcube_t
            ):
```

src/fmu/tools/domainconversion/dconvert.py
```diff
@@ -359,6 +389,28 @@ def _extend_incube_create_speed_cube_average(
...
+                _logger.debug("Reuse velocity cube for time->depth conversion")
+                _logger.debug("Update internal time cube")
+                self._time_cube = _incube
+                return
+        else:
+            if (
+                hasattr(self, "_scube_d")
+                and bool(self._scube_d)
+                and self._same_cube_geometry(_incube, self._scube_d)
+            ):
```
tnatt: Same comment as above

src/fmu/tools/domainconversion/dconvert.py
anbec12: Agree

src/fmu/tools/domainconversion/dconvert.py
anbec12: It cannot happen: 'ncol', 'nrow', 'nlay', 'xinc', 'yinc', and 'zinc' are required inputs to a cube and the other attributes are assigned defaults.

src/fmu/tools/domainconversion/dconvert.py
anbec12: Agree

anbec12: I have updated the function to this now, I think this is even more readable:
```    
    @staticmethod
    def _same_cube_geometry(cube1: xtgeo.Cube, cube2: xtgeo.Cube) -> bool:
        """Check if two cubes have the same geometry."""
        keys = ("xori", "yori", "zori", "xinc", "yinc", "zinc", "rotation", "yflip")

        if cube1.dimensions != cube2.dimensions:
            return False

        for k in keys:
            if k in ["xori", "yori"]:
                atol = 1e-2
            else:
                atol = 1e-6

            v1, v2 = float(getattr(cube1, k)), float(getattr(cube2, k))
            if not np.isclose(v1, v2, atol=atol):
                return False

        return True
```

src/fmu/tools/domainconversion/dconvert.py
anbec12: Awesome, much better with your suggestion!

src/fmu/tools/domainconversion/dconvert.py
anbec12: Also same comment as above

src/fmu/tools/domainconversion/dconvert.py
tnatt: Yes, looks neat 😃  maybe even do an inline for the `atol`..but this is just a matter of taste 
`atol = 1e-2 if k in ["xori", "yori"] else 1e-6`

src/fmu/tools/domainconversion/dconvert.py
tnatt: Ok great, thanks for claring that up!

tnatt (approved):
Nice improvement 👍

PR #455, opened 2026-06-24
ENH: Efficient inter- and extrapolation of speedcube

### LINKED ISSUE #446: Flaws in domainconversion _extend_incube_create_speed_cube_average

anbec12:
Resolves #446 issue #2 within

Changed from two loops to loop over each trace, which could be up to a million, to one loop over depth indices, which is typically only a few hundreds at most. The speed down to the shallowest point on the first surface below MSL is assigned in one step as it is constant.

These improvements reduce the time needed for interpolation of the speedcube enormously.

## Checklist

- [ ] Docstrings are correct and updated

### DIFF
--- a/src/fmu/tools/domainconversion/dconvert.py
+++ b/src/fmu/tools/domainconversion/dconvert.py
@@ -430,9 +430,12 @@ class DomainConversion:
         _logger.debug("Create average %s cube", result_description)
         speedcube = cube.copy()
 
-        cube_arr = [cube.zori + n * cube.zinc for n in range(cube.values.shape[2])]
-
-        speedcube_values = np.zeros_like(cube.values)
+        # Use float32 where possible to increase speed
+        ni, nj, n = cube.values.shape
+        z0 = np.float32(cube.zori)
+        zinc = np.float32(cube.zinc)
+        cube_arr = z0 + zinc * np.arange(n, dtype=np.float32)
+        speedcube_values = np.zeros_like(cube.values, dtype=np.float32)
 
         # Resample surfaces to match the input cube
         x_surfaces, y_surfaces = self._resample_surfaces_to_cube(cube, time2depth)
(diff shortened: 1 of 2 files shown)


src/fmu/tools/domainconversion/dconvert.py
```diff
@@ -402,37 +402,84 @@ def _extend_incube_create_speed_cube_average(
...
+            # Right side (x0 > X[-1]): linear extrapolation using last slope
+            mask_right = x0 > X[-1]
+            if mask_right.any():
+                yh_last = Y[-1][mask_right]
+                xh_last = X[-1][mask_right]
+                sl_last = S_last[mask_right]
+                y2d[mask_right] = yh_last + sl_last * (x0 - xh_last)
 
-        # Make sure first sample also gets a non-zero value
-        speedcube_values[:, :, 0] = speedcube_values[:, :, 1]
```
GibranAlfa, line 463: I think we should keep the old handling of the first slice here.

Could we add that back after the new loop?

`speedcube_values[:, :, 0] = speedcube_values[:, :, 1]`

That should keep the new faster approach, but preserve the old behavior for the top sample.

anbec12: I have removed that line on purpose because the top sample `speedcube_values[:, :, 0]` is assigned further up in the code. In the old code, the top sample was not assigned in the loops and, therefore, this line was necessary after the loops. 
 ```
 # OLD code ----
 # Loops
 for i in range(speedcube_values.shape[0]):
            for j in range(speedcube_values.shape[1]):
                ...
                if time2depth:
                    speedcube_values[i, j, 1:] = y_arr[1:] / cube_arr[1:] * 2000
                else:
                    speedcube_values[i, j, 1:] = y_arr[1:] / cube_arr[1:] / 2000

# After loops - requires assignment of top sample
speedcube_values[:, :, 0] = speedcube_values[:, :, 1]

# NEW code - top sample is assigned already ----
 if time2depth:
    speedcube_values[:, :, :k1_min] = S0[:, :, None] * 2000.0
else:
    speedcube_values[:, :, :k1_min] = S0[:, :, None] / 2000.0
```

Comment (GibranAlfa) [reply src/fmu/tools/domainconversion/dconvert.py:463]: Yes, I agree that slice 0 is assigned now, so it is not the same as the old “unassigned first sample” case.

My concern is the value it gets when MSL is already present in the input surfaces. We always insert an internal MSL surface, so an input like [0, 10, 100] becomes [0, 0, 10, 100] internally. In that case `S0` is based on the duplicated 0 surface, not the first real interval below MSL.

The existing MSL test still passes because it gives `zinc=1.0` explicitly. The issue shows up when `zinc` is left to be estimated from the speed cube, since the median speed is then used for the output spacing.

So I think we should either copy slice 1 into slice 0 like before, or make the shallow assignment use the first real interval below MSL.

Comment (anbec12) [reply src/fmu/tools/domainconversion/dconvert.py:463]: This cannot happen anymore. I have addressed the issue in PR #448. 

In the old code, if MSL was given as input surfaces, you could get average velocities of ~0 m/s in the first interval __MSL to MSL. This velocity was assigned to the MSL surface and copied to __MSL. This could lead to stretching artifacts in the domain converted cube as the velocity from MSL (~0 m/s) was interpolated linearly to the average velocity of the first real subsurface surface (~2000 m/s in Drogon). 

After PR #448, the velocity (or slowness) surfaces above the real first subsurface surface, i.e.  __MSL and MSL if given, are always the same as the velocity down to the real first subsurface surface.

If I understand your suggestions in the last sentence correctly, than the PR #448 has already implemented that.

Comment (GibranAlfa) [reply src/fmu/tools/domainconversion/dconvert.py:463]: It is not a problem for the current `smallcube` case, because that fixture has 101 vertical layers and one bad first slice does not move the median. But it can be a problem for shallow cubes, or any case where the first slice has enough weight in the median used for default `zinc`.

```py
def test_speedcube_with_input_msl_and_default_zinc() -> None:
    cube = xtgeo.Cube(
        ncol=1,
        nrow=1,
        nlay=2,
        xinc=1.0,
        yinc=1.0,
        zinc=1.0,
        values=np.zeros((1, 1, 2)),
    )
    surface_template = xtgeo.surface_from_cube(cube, value=0)

    d0 = surface_template.copy()
    d0.values = 0.0
    d1 = surface_template.copy()
    d1.values = 10.0

    dc = DomainConversion([d0, d1], [d0, d1])

    result = dc.depth_convert_cube(cube, zmin=0, zmax=10)  # zinc omitted on purpose

    assert result.zinc == pytest.approx(1.0)
    assert dc.average_velocity_cube_in_time.values[0, 0, 0] == pytest.approx(2000.0)
```

First assertion would fail with:
```text
comparison failed
Obtained: 0.5
Expected: 1.0 ± 1.0e-06
```
If that assertion is commented/reordered, the second one also shows that the first speed slice got the fallback value:
```text
comparison failed
Obtained: 1.0
Expected: 2000.0 ± 0.002
```
This probably does not affect normal larger cubes much, since the median still ends up correct. But it is still a reachable edge case when `zinc` is estimated and the cube has few vertical samples. In that case the fallback value in slice 0 changes the median speed and the inferred output `zinc`. You can decide if this is worth addressing or not.

anbec12: Well spotted! I added the line `speedcube_values[:, :, 0] = speedcube_values[:, :, 1]` to `dconvert.py`. I also added the test you have written above to `test_domainconversion.py`.

This solves the issue you brought up.

GibranAlfa (approved):
looks good to me 😊

PR #457, opened 2026-06-25
ENH: Add calculation of interval velocity maps

### LINKED ISSUE #456: Add interval velocity maps to domainconversion
Interval velocity maps are normal QC products for velocity model evaluation.
Add calculation of interval velocity maps to the DomainConversion object to facilitate easy access e.g. in RMS scripts.
Currently only average velocity and slowness maps are 'offered', but they are not very useful to evaluate the model.

anbec12:
Resolves #456 

Add internal function `_calculate_interval_velocity_surfaces`. This is called in the `_check_fix_surfaces` function. A public generator method `interval_velocity_surfaces` is also added to give 'public' access to the surfaces. I think the `_calculate_interval_velocity_surfaces` function could also be useful for potential future features, like velocity model smoothing or clamping velocities to reasonable (physical) limits.

I see that names as well as velocity and slowness surfaces are accessible via generators instead of returning just lists. I am not sure why generators are preferred over lists for these relatively small items, but I think generators are perhaps a little bit less user-friendly as you have to first generate a list if you need the items in a list. Anyway, I also use a generator  in the `interval_velocity_surfaces` method for consistentcy.

I have not updated documentation. Don't know how to do that.

## Checklist

- [ ] Documentation is updated, if necessary  (necessary, but not done)

### DIFF
--- a/src/fmu/tools/domainconversion/dconvert.py
+++ b/src/fmu/tools/domainconversion/dconvert.py
@@ -60,6 +60,7 @@ class DomainConversion:
     _scube_d: xtgeo.Cube = field(default=None, init=False)  # avg slowness cube in depth
 
     _v_surfaces: list[xtgeo.RegularSurface] = field(default_factory=list, init=False)
+    _vi_surfaces: list[xtgeo.RegularSurface] = field(default_factory=list, init=False)
     _s_surfaces: list[xtgeo.RegularSurface] = field(default_factory=list, init=False)
     _d_surfaces: list[xtgeo.RegularSurface] = field(default_factory=list, init=False)
     _t_surfaces: list[xtgeo.RegularSurface] = field(default_factory=list, init=False)
@@ -121,6 +122,7 @@ class DomainConversion:
             self.time_surfaces, template_surf, fill=True, ensure_consistency=True
         )
         self._ensure_surfaces_has_msl()
+        self._calculate_interval_velocity_surfaces()
         _logger.debug("Check and fix surfaces... DONE")
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


src/fmu/tools/domainconversion/dconvert.py
```diff
@@ -706,6 +734,11 @@ def slowness_surfaces(self) -> Generator[xtgeo.RegularSurface]:
         for surf in self._s_surfaces:
             yield surf
 
+    def vint_surfaces(self) -> Generator[xtgeo.RegularSurface]:
+        """Return a generator of interval velocity surfaces."""
+        for surf in self._vi_surfaces:
+            yield surf
+
```
tnatt, line 741: I do question whether any of these `surface_names, velocity_surfaces, slowness_surfaces and vint_surfaces` should be returned as generators.

Since the lists they are serving have already been made upfront they are already stored in memory, a generator is not more memory-efficient. Also users are more custom to working with lists than generators, for many operations now (indexing etc) they will have to convert it to list first.

Do we know if `DomainConversion` is in use in user scripts or is it mainly sim2seis that uses it?

I don't think we should change it in this PR, but WDYT should we make an issue to change these properties?

```python
    def vint_surfaces(self) -> list[xtgeo.RegularSurface]:
        """Return the list of interval velocity surfaces."""
        return self._vi_surfaces
```

tnatt (approved):
LGTM

src/fmu/tools/domainconversion/dconvert.py
anbec12: I would prefer lists to be returned. We can make an issue

PR #464, opened 2026-07-01
ENH: Add NestedHybridGrid class

### LINKED ISSUE #461: Make NestedHybridGrid into a class
Currently users create a nested hybrid grid through a function `create_nested_hybrid_grid()`.
This function returns two object, the `grid` and the `nnc_table` with cell connections. 

In #439 a change is introduced that makes this `create_nested_hybrid_grid()` return three objects. This is not the ideal.

Switching to a class instead of a function is better as it will make the interface clearer, easier to use, and more maintainable. Users interact with a single object instead of multiple return values, which reduces errors and supports future extensions.

Examples of what a`NestedHybridGrid` class could contain

**Public attributes** 
- grid
- nnc_table
- properties

**Public methods**
- to_rms()
- write_nnc_table()
- modify_upscaling_mapping()

**Class methods**
- from_rms()

It should be easy to use this from within RMS, as most users will construct the nested hybrid grid there.
From RMS it could looks something like:

```python
gridname = "Simgrid"
properties = ["Regions"]
refinement = (2, 2, 2)
region = "nested_region"
nested_gridname = "Nestedgrid"
nnc_filename = "nnc_file.csv"


def create_nested_hybrid_grid():

    nhg = NestedHybridGrid.from_rms(project, gridname, refinement, region, properties)
    nhg.write_nnc_table(nnc_filename)

    # write grid and included properties back to rms (includes zone)
    nhg.to_rms(project, nested_gridname)

If __name__=="__main__":
     create_nested_hybrid_grid()

```

tnatt:
Resolves #461 

First initial PR to introduce a `NestedHybridGrid` class, to allow for future extension and ease the maintenance.
It contains three public properties:

- `grid` - which is the final merged grid
- `nnc_table` - which is the nnc table with cell connections between the refined and coarse area
- `properties` - which is the list of the attached grid properties 

Mostly code has just been moved to fit into a class design, and the logic stays the same. 

The `create_nested_hybrid_grid()` still exists but uses the `NestedHybridGrid` in the background. 

There are plans to expand the class with a simple to use `from_rms()` and `to_rms` methods. Once they are in place, the documentation will be updated.

## Checklist


src/fmu/tools/nestedhybridgrid/nestedhybrid.py
```diff
@@ -36,58 +37,76 @@
...
     """
-    filled = np.ma.filled(region_prop.values, fill_value=-1).astype(int)
-    ni, nj, nk = filled.shape
+    ni, nj, nk = refinement_area.shape
+    active = ~refinement_area.mask
 
-    in_target = filled == target_region
-    outside_active = (filled != target_region) & (filled != -1)
+    in_target = refinement_area & active
+    outside_active = ~refinement_area & active
```
tnatt: The original code just filled masked values with -1, but then those -1 was excluded from the `in_target` and `outside_active` filters. Hence the filling is unnecessary and the `mask` can be used directly to define the filters.

src/fmu/tools/nestedhybridgrid/nestedhybrid.py
```diff
@@ -121,15 +140,13 @@ def _find_boundary_faces(
...
+    _logger.info("Found %d boundary faces for refinement area", len(faces))
     return faces
 
 
 def _compute_nnc_table(
-    region_prop: xtgeo.GridProperty,
-    target_region_id: int,
-    crop_origin: tuple[int, int, int],
-    refinement: tuple[int, int, int],
+    refinement_area: np.ma.MaskedArray,
```
tnatt: In general this PR is mainly just a move of the original code into a class design, this is however a place where a change was made, By entering the `refinement_area` instead we can reduce the number of arguments in this function.

src/fmu/tools/nestedhybridgrid/nestedhybrid.py
```diff
@@ -255,86 +273,193 @@ def _compute_nnc_table(
...
 
-def _set_actnum_by_region(
-    grid: xtgeo.Grid,
-    region_prop: xtgeo.GridProperty,
-    target_region: int,
-    *,
-    invert: bool = False,
-) -> None:
-    """Deactivate cells based on region membership."""
+def _set_actnum_in_grid(grid: xtgeo.Grid, active_mask: np.ndarray) -> None:
```
tnatt, line 278: This function was simplified and the invert logic was removed.

src/fmu/tools/nestedhybridgrid/nestedhybrid.py
```diff
@@ -255,86 +273,193 @@ def _compute_nnc_table(
...
+        return _compute_nnc_table(
+            refinement_area=self._refinement_area,
+            refinement=self._refinement,
+            coarse_ncol=self._original_dimensions.ncol,
+            lmap1=self._layer_map_coarse,
+            lmap2=self._layer_map_refined,
+        )
 
-    updated_subgrid = {}
+    def _generate_layer_map_coarse(self) -> np.ndarray:
```
tnatt, line 416: The layer mapping logic is exactly as before, just split `_generate_layer_mappings` into two, one for the coarse and one for the refined area.

src/fmu/tools/nestedhybridgrid/nestedhybrid.py
```diff
@@ -255,86 +273,193 @@ def _compute_nnc_table(
...
+            layer_map1=self._layer_map_coarse,
+            layer_map2=self._layer_map_refined,
+        )
+        _logger.info("Merged grid dimensions: %s", grid.dimensions)
+
+        grid.subgrids = self._set_zonation(grid.nlay)
+        return grid
+
+    @property
+    def grid(self) -> xtgeo.Grid:
```
slangeveld, line 388: Same as above, should we rename this to `nested_grid` to reflect what it is?

src/fmu/tools/nestedhybridgrid/nestedhybrid.py
```diff
@@ -386,71 +511,15 @@ def create_nested_hybrid_grid(
         "breaking changes in future versions without notice.",
         FutureWarning,
     )
-    if any(r < 1 for r in refinement):
```
slangeveld, line 389: Will be a nice improvement to reduce the length and logic of this method like you have done here 👍

tests/nestedhybridgrid/test_nestedhybrid.py
```diff
@@ -230,6 +287,276 @@ def test_zonation_with_layer_offset(self):
...
+        assert _has_single_row(i1=1, j1=2, k1=1, i2=5, j2=1, k2=1, direction="I+")
+        # bottom neighbor → connects to J- face of refined patch
+        assert _has_single_row(i1=2, j1=1, k1=1, i2=6, j2=1, k2=1, direction="J+")
+        # top neighbor → connects to J+ face of refined patch
+        assert _has_single_row(i1=2, j1=3, k1=1, i2=5, j2=2, k2=1, direction="J-")
+        # right neighbor → connects to I+ face of refined patch
+        assert _has_single_row(i1=3, j1=2, k1=1, i2=6, j2=2, k2=1, direction="I-")
+
+    def test_inactive_neighbor_excluded_from_nnc(self):
+        """Inactive mother cells should not appear in the NNC table.
```
slangeveld: Mother or neighbor cells? Both are used here. Probably good to choose one and be consequent

src/fmu/tools/nestedhybridgrid/nestedhybrid.py
```diff
@@ -255,86 +273,193 @@ def _compute_nnc_table(
...
-        (rlay - 1) * np.minimum(int(refined_nlay / rlay), lmap1 - k0),
-    )
-    lmap2 = np.arange(refined_nlay, dtype=np.int32) + k0
+        if not isinstance(refinement, tuple) or len(refinement) != 3:
+            raise ValueError(
+                "Refinement must be a tuple of three integers: (rcol, rrow, rlay)"
+            )
 
-    return (lmap1, lmap2)
+    def _build_grid(self) -> xtgeo.Grid:
```
slangeveld: Should the method be named `_build_nested_grid()` instead to clearify that this is not a normal grid?

slangeveld (approved):
Looks good to me 👍  A few comments to consider

src/fmu/tools/nestedhybridgrid/nestedhybrid.py
tnatt: Yes agree with that! I updated the name of this method to `_build_nested_hybrid_grid()` 🙂

src/fmu/tools/nestedhybridgrid/nestedhybrid.py
tnatt: Not sure.. I feel the class name provides enough context, so I decided to keep the simple `grid` property name. I did however rename the input argument in the class from `grid` to `coarse_grid` to reduce ambiguity 🙂

tests/nestedhybridgrid/test_nestedhybrid.py
tnatt: agree, I went with coarse neighbor 🙂  not sure I like the term mother.

PR #490, opened 2026-08-12
ENH: Add to_rms()/from_rms() methods in NestedHybridGrid

### LINKED ISSUE #468: Add `from_rms()` / `to_rms()` methods on `NestedHybridGrid`
It should be easy to use the `NestedHybridGrid` class from within RMS, as most users will construct the nested hybrid grid there.
We should add a `from_rms()` class method to easily construct an instance of the class, and a `to_rms()` methods which should store the grid and all appended properties as a grid model inside RMS.

From RMS it could look something like:

```python
gridname = "Simgrid"
properties = ["Regions"]
refinement = (2, 2, 2)
region = "nested_region"
nested_gridname = "Nestedgrid"
nnc_filename = "nnc_file.csv"


def create_nested_hybrid_grid():

    nhg = NestedHybridGrid.from_rms(project, gridname, refinement, region, properties)
    nhg.write_nnc_table(nnc_filename)

    # write grid and included properties back to rms (includes zone)
    nhg.to_rms(project, nested_gridname)

If __name__=="__main__":
     create_nested_hybrid_grid()

```

tnatt:
Resolves #468 

PR that simplifies the initialization of the `NestedHybridGrid` in RMS, by allowing users to enter names of objects inside RMS instead of xtgeo objects. It also enables easy writing of the grid and all of it's appended properties to RMS.  

## Checklist


tests/nestedhybridgrid/test_nestedhybrid.py
alifbe: `_make_box_grid_with_region` by default  create region with id = 2, which means the test below should not create any nested grid. But since the assertion is only checking mock  call counts/arguments, this didn't get caught.

alifbe (changes requested):
The new class is expecting region value = 1 for nested hybrid grid but the coarse grid is created with region value = 2. But none of the test capture this bug.

Suggest to change test strategy to check if nested grid is created.

tnatt:
> The new class is expecting region value = 1 for nested hybrid grid but the coarse grid is created with region value = 2. But none of the test capture this bug.
> 
> Suggest to change test strategy to check if nested grid is created.

Good catch 👀 will update!

tests/nestedhybridgrid/test_nestedhybrid.py
tnatt: the `_make_box_grid_with_region` creates a region property where half the values are 1 and half the values are 2... so we are creating a nested grid - just refining the other half than what the target_region_id is saying .. a bit confusing 😄  so will make it more clear and add more test that we do refine the area that has value 1

tnatt:
Have a look again @alifbe, I added more checks regarding the actual output itself - not just if expected calls are being made

src/fmu/tools/nestedhybridgrid/nestedhybrid.py
alifbe: Not related, but this should be `list[xtgeo.GridProperty]` right?

alifbe (approved):
This looks great 💯 

Just spot something unrelated to this PR. Not sure if it can be part of this PR or another PR.

src/fmu/tools/nestedhybridgrid/nestedhybrid.py
tnatt: yup! will fix it while here

PR #491, opened 2026-08-13
ENH: Add method to write nnc table to file

### LINKED ISSUE #469: Add method in `NestedHybridGrid` to write the nnc table to file
The `nnc_table` is a `pandas.DataFrame` which is stored as a property on the `NestedHybridGrid` class.

Users need to save the `nnc_table` to disk, as it will be required again after property upscaling to generate the final NNC table for the simulator.

Currently they need to do:

```python
nhg = NestedHybridGrid(..)
nhg.nnc_table.to_csv(filename, index=False)
```

it would be nice with a convenience method that wraps the above e.g.

```python
nhg = NestedHybridGrid(..)
nhg.write_nnc_table(filename)
```

nourinmohd:
Resolves #469 

Implemented a convenience method for writing the nested hybrid grid NNC table to disk.

## Checklist

- [ ] Tests added (if not, comment why)
- [x] Added comments on this PR where appropriate to help reviewers

### DIFF
--- a/src/fmu/tools/nestedhybridgrid/nestedhybrid.py
+++ b/src/fmu/tools/nestedhybridgrid/nestedhybrid.py
@@ -430,6 +430,10 @@ class NestedHybridGrid:
             self._nnc_table = self._compute_nnc_table()
         return self._nnc_table
 
+    def write_nnc_table(self, filename: str | os.PathLike[str]) -> None:
+        """Write the NNC mapping table to CSV."""
+        self.nnc_table.to_csv(filename, index=False)
+
     def _compute_nnc_table(self) -> pd.DataFrame:
         """Compute the NNC mapping table."""
         return _compute_nnc_table(
--- a/src/fmu/tools/nestedhybridgrid/nestedhybrid.py
+++ b/src/fmu/tools/nestedhybridgrid/nestedhybrid.py
@@ -18,6 +18,7 @@ from __future__ import annotations
 
 import logging
 import warnings
+from pathlib import Path
 from typing import TYPE_CHECKING, Any, Literal, Self, TypeAlias
 
 import numpy as np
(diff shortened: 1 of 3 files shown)


src/fmu/tools/nestedhybridgrid/nestedhybrid.py
```diff
@@ -403,6 +403,10 @@ def nnc_table(self) -> pd.DataFrame:
             self._nnc_table = self._compute_nnc_table()
         return self._nnc_table
 
+    def write_nnc_table(self, filename: str | os.PathLike[str]) -> None:
+        """Write the NNC mapping table to CSV."""
+        self.nnc_table.to_csv(filename, index=False)
```
tnatt: Suggest using Path here which is more modern 👍 you need to import it from `pathlib` if we don't already.
And I think we should ensure that output folders are created if they don't exists 🙂 You could add an extra test for that.

```suggestion
    def write_nnc_table(self, filename: str | Path) -> None:
        """Write the NNC mapping table to CSV."""
        output_path = Path(filename)
        output_path.parent.mkdir(parents=True, exist_ok=True)
        self.nnc_table.to_csv(output_path, index=False)
```

nourinmohd: - Updated `nestedhybrid.py` to use `Path`, create missing parent directories, and write via the resulting path.
- Added a regression test for nested output directories in `test_nestedhybrid.py`.

nourinmohd:
- Add a test that output folders are created if they don't exist.

tests/nestedhybridgrid/test_nestedhybrid.py
```diff
@@ -443,6 +443,34 @@ def _has_single_row(i1, j1, k1, i2, j2, k2, direction):
         # right neighbor → connects to I+ face of refined patch
         assert _has_single_row(i1=3, j1=2, k1=1, i2=6, j2=2, k2=1, direction="I-")
 
+    def test_write_nnc_table_writes_csv_without_index(self, tmp_path):
+        """NNC table can be written to CSV without the pandas index."""
+        grid = xtgeo.create_box_grid((3, 3, 1))
+
+        region = xtgeo.GridProperty(grid, name="REGION", discrete=True, values=0)
+        region.values[1, 1, 0] = 1
+
```
tnatt: You could utilize the existing `_make_box_grid_with_region` to reduce some code in your tests

```suggestion
        grid, region, _ = _make_box_grid_with_region(dimension=(3, 3, 1))
```

tnatt (commented):
Just remembered, could you update the documentation to use the new method also 🙂 ? In the quick start example

nourinmohd:
- The output-directory test verifies nested parent folders are created before writing: `test_nestedhybrid`
- Both NNC writing tests now reuse `_make_box_grid_with_region`.
- The quick-start documentation uses `nhg.write_nnc_table(...`): `nestedhybridgrid.rst`.

tnatt (approved):
Nice :)

PR #492, opened 2026-08-13
ENH: Add original I,J,K as properties in NestedHybridGrid

### LINKED ISSUE #470: Add original I,J,K as properties in the `NestedHybridGrid`
The original I, J, K values (row, col, layer) should be preserved and added as properties on the `NestedHybridGrid` class.

There are multiple use cases where the original IJK is necessary to know, e.g. for post-processing purposes. These properties could easily be transferred from the original input grid to the nested grid by appending them as properties to the coarse grid before the grid merge is run. This will transfer the property automatically to the nested grid by the `grid_merge()` functionality.

In the `_build_nested_hybrid_grid()` we could after appending the input region add these lines:

```python
for prop in coarse_grid.get_ijk(names=("I_orig", "J_orig", "K_orig")):
    coarse_grid.append_prop(prop)

```
This will append I,J,K to the grid with the names  `I_orig,` `J_orig,` and `K_orig`. They will then appear and be available in the `properties` property and can also be fetched from the final nested hybrid grid like so  `nhg.grid.get_prop_by_name("I_orig")`

nourinmohd:
Resolves #470 

- Adds preservation of original grid I/J/K coordinates for `NestedHybridGrid`.

- `_build_nested_hybrid_grid()` now appends `I_orig, J_orig`, and `K_orig` properties to the coarse grid before refinement and merge, allowing `xtgeo.grid_merge() `to transfer them automatically to the final nested hybrid grid. These properties are available through `nhg.properties` and `nhg.grid.get_prop_by_name(...)`.

- Also updates the nested hybrid grid property test to verify the new properties are present and retain the original 1-based coordinates.

## Checklist

- [ ] Tests added (if not, comment why)
- [x] Added comments on this PR where appropriate to help reviewers


nourinmohd:
- Add a constant for the I_org, J_org, and K_org instead of passing them as strings.

src/fmu/tools/nestedhybridgrid/nestedhybrid.py
```diff
@@ -40,6 +40,7 @@
 FaceDirection: TypeAlias = Literal["i+", "i-", "j+", "j-", "k+", "k-"]
 IJKIndex: TypeAlias = tuple[int, int, int]
 BoundaryCellFace: TypeAlias = tuple[IJKIndex, IJKIndex, FaceDirection]
+ORIGINAL_IJK_PROPERTY_NAMES = ("I_orig", "J_orig", "K_orig")
```
alifbe: We discussed a bit and thinking that perhaps the property name can be `parent_I`, `parent_J` and `parent_K`

tests/nestedhybridgrid/test_nestedhybrid.py
```diff
@@ -380,17 +384,33 @@ def test_nestedhybridgrid_properties_as_expected(self):
         nhg = NestedHybridGrid(coarse_grid=grid, region=region, refinement=(2, 2, 2))
 
         # check that the region property is attached to the merged grid
-        assert len(nhg.properties) == 2
+        assert len(nhg.properties) == 5
 
         prop_names = [prop.name for prop in nhg.properties]
-        assert set(prop_names) == {"CUSTOM", "REGION"}
+        assert set(prop_names) == {"CUSTOM", "REGION", "I_orig", "J_orig", "K_orig"}
```
alifbe: Since you already imported the property names constant, perhaps you can use it here.

tests/nestedhybridgrid/test_nestedhybrid.py
```diff
@@ -380,17 +384,33 @@ def test_nestedhybridgrid_properties_as_expected(self):
...
-        region_prop = nhg.properties[0]
+        region_prop = nhg.grid.get_prop_by_name("REGION")
 
         assert isinstance(region_prop, xtgeo.GridProperty)
         assert region_prop.name == "REGION"
         assert region_prop.values.shape == nhg.grid.dimensions
 
+        i_orig = nhg.grid.get_prop_by_name("I_orig")
+        j_orig = nhg.grid.get_prop_by_name("J_orig")
+        k_orig = nhg.grid.get_prop_by_name("K_orig")
```
alifbe: Similar as above.

alifbe (changes requested):
These properties is very useful for mapping between nested grid and parent grid which is hopefully can be fully automated from within NHG class. So, at least for now, we don't see the needs for this property to be exported to RMS. Please ensure that `.to_rms()` method doesn't export these IJK properties.

nourinmohd:
Implemented:

- Renamed IJK properties to `parent_I, parent_J, and parent_K`.
- `.to_rms()` now excludes these internal mapping properties.
- Updated tests to use `ORIGINAL_IJK_PROPERTY_NAMES`.

tnatt (approved):
This looks good to me 🙂 
We might want to add some info regarding these properties into the documentation, to inform they are there.. but we can maybe wait and see if the need arises. If there is a need, the best solution would perhaps be to add a flag in the `to_rms` function to easily store these properties in rms.

alifbe (approved):
Looks great. Please consider changing the commit message. For some case where user do nested over nested, original could be interpreted as origin grid. But in this case, the property that is created is direct parent.

PR #494, opened 2026-08-14
FIX: Only emit FutureWarning when method is not given

### LINKED ISSUE #493: Make `FutureWarning` in `DomainConversion` less eager
We have a `FutureWarning` in `DomainConversion` that tells users that the default value of an argument will change. 
This warning is emitted always, but should instead be emitted only if the argument was not provided. 

https://github.com/equinor/fmu-tools/blob/e344fe9b5b504167b18b557ffebf6241629060d2/src/fmu/tools/domainconversion/dconvert.py#L967-L973

tnatt:
Resolves #493 

## Checklist

- [ ] Docstrings are correct and updated


rnyb (approved):
LGTM

PR #504, opened 2026-08-31
ENH: Support OWC in create_fluid_contacts_from_grid

### LINKED ISSUE #503: Support OWC as contact type in create_fluid_contacts_from_grid
`OWC` was by choice kept out as a contact type in `create_fluid_contacts_from_grid` - if you have an `OWC` parameter in the grid, it is most likely representing the `FWL` as `OWC` is difficult to estimate and will vary based on the properties in the cells and height above FWL.

However some users have reported that it is a bit confusing due to several reasons

- We support GWC, isn't that also an FWL?
- how to differentiate an area with only gas from an area with both oil and gas
- OWC is a supported contact type in the simple exports in fmu-dataio

We should probably just support OWC, and then it is up to the user to make the best call for what to name their contacts so it makes sense in the visualisation.

tnatt:
Resolves #503 

## Checklist


nourinmohd (approved):
LGTM 👍

tests/rms/test_create_fluid_contacts_from_grid.py
```diff
@@ -272,22 +272,38 @@ def test_rms_data_loader_find_available_contact_properties(
     mock_loader.project.grid_models["mygrid"] = MagicMock(properties=["myfwl", "mygwc"])
 
     # GWC is not present in the grid model
-    contacts = mock_loader.find_available_contact_properties("myfwl", "GOC", "mygwc")
+    contacts = mock_loader.find_available_contact_properties(
+        "myfwl", "GOC", "mygwc", "OWC"
+    )
     assert len(contacts) == 2
```
nourinmohd: Just curious, do we not assert for OWC here as well ?

krhart (approved):
Looks good 👍
(I can't recall why we didn't include it in the first place, but I remember we had a lot of discussions about how to do it 😅)

tnatt:
> Looks good 👍 (I can't recall why we didn't include it in the first place, but I remember we had a lot of discussions about how to do it 😅)

Yes I also remember we discussed this thoroughly 😄

